PR #6, opened 2018-12-17
[WIP] Change how auto_install plugins are added

martin308:
Changes the way that plugins are auto installed. By using strings to identify the plugins and check for their presence we can alter the sequel plugin to only define it's modules if required. We need to do this as defining the Sequel constant when not necessary is causing issues with other gems.


qnm:
I believe I have the reverse of this issue. It looks like beelines causes part of ActiveRecord to be loaded, which confuses a gem which relies on autodetection (Fabrication, fwiw).

I will test this and let you know.

PR #16, opened 2019-04-12
Allow calling #span without a block

eternal44:
I'd like to add new event fields to a span without having to pass in an empty block:

`Honeycomb.span(type: 'worker', fields: some_hash) {}`

I didn't see anything in the readme regarding PR guidelines. This would be a nice-to-have but not an urgent change.

### DIFF
--- a/lib/honeycomb/span.rb
+++ b/lib/honeycomb/span.rb
@@ -44,7 +44,7 @@ module Honeycomb
 
         start = Time.now
 
-        yield span_id, trace_id
+        yield span_id, trace_id if block_given?
       end
     rescue Exception => e
       if event


martin308:
@eternal44 could you explain a bit more of your use-case here? If you call it without the block then the span will be sent straight away so your duration_ms value on the span will be mostly useless

eternal44:
@martin308 My apologies thanks for clarifying, I want to to add additional fields to a span and this is probably the wrong way to do that. I think I should be using [#add](https://github.com/honeycombio/libhoney-rb/blob/master/lib/libhoney/client.rb#L154) instead?

martin308:
@eternal44 No worries, if you have a reference to the current span then you can call add. That can be a little tricky with the current implementation. Reach out to support@honeycomb.io and we can help further with getting your tracing setup!

PR #26, opened 2019-09-01
Set sample hook on child spans

orangejulius:
I have an application which has a health check endpoint, which I'd like to filter out from Honeycomb. The endpoint gets a lot of requests and they aren't particularly interesting to log.

While trying to set a sample_hook, I noticed that child spans created by the ActiveSupport subscriber integration aren't being filtered out. Since every request to this healthcheck endpoint was generating a child span via ActionController, the sample hook was only able to filter out half of the spans. 

This PR sets the `sample_hook` parameter when creating child spans, so that any span can potentially be filtered out.

Let me know if there's a better way to accomplish this. I noticed that [`send_children`](https://github.com/honeycombio/beeline-ruby/blob/master/lib/honeycomb/span.rb#L101) is called in `span.rb` before testing whether the span should be sampled. So it looks like another approach would be to move that call to a little later in the `send_internal` method.

### DIFF
--- a/lib/honeycomb/span.rb
+++ b/lib/honeycomb/span.rb
@@ -51,7 +51,8 @@ module Honeycomb
       self.class.new(trace: trace,
                      builder: builder,
                      context: context,
-                     parent_id: id).tap do |c|
+                     parent_id: id,
+                     sample_hook: @sample_hook).tap do |c|
         children << c
       end
     end
--- a/lib/honeycomb/span.rb
+++ b/lib/honeycomb/span.rb
@@ -52,7 +52,8 @@ module Honeycomb
                      builder: builder,
                      context: context,
                      parent_id: id,
-                     sample_hook: @sample_hook).tap do |c|
+                     sample_hook: sample_hook,
+                     presend_hook: presend_hook).tap do |c|
         children << c
       end
     end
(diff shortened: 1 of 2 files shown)


martin308:
@orangejulius thanks for the PR! I'll just update the test and add in passing through the `presend_hook` at the same time

orangejulius:
Sure thing. Also, I didn't realize that at the last minute i made a change to the PR that broke the tests, so I just pushed a fix.

How do you feel about using `instance_variable_get` in the tests to fetch the expected `sample_hook`?

martin308:
@orangejulius ah I have some commits against your old commits that fixes the test without using the `instance_variable_get` on sample hook. Mind if I force push the commits?

orangejulius:
Go for it, thanks!

martin308:
@orangejulius thanks, sorry about the confusion! Let me know what you think of what I just pushed 😄

orangejulius:
Looks great. Using test doubles and asserting correct behavior rather than whether or not certain instance variables are set is far better :)

Also, wow you have a lot of Travis tests!

martin308:
@orangejulius thanks for looking it over! Yeh they caught a few things across ruby versions and rails versions while I was building it so they are helpful but they take a long time to run! 😆

Issue #25, opened 2019-08-14
start_span delegates to nil if `configure` hasn't run


DanielHeath:
### Description
Upgraded to 1.0, ran my tests.

### Issue
Loads of test failures - `undefined method `start_span' for nil:NilClass`

Looks like @client is nil in my test environment, since I don't want to send anything.

Maybe `start_span` should be a silent no-op if no client is configured?

Sergio-Mira:
Seeing the same issue on my test environment

martin308:
Hopefully this is fixed with #27, it's being released in `v1.0.1` of the beeline. Feel free to reopen this issue if you're still seeing this after upgrading!

PR #27, opened 2019-09-03
No-op if no client found in Faraday integration

Sergio-Mira:
https://github.com/honeycombio/beeline-ruby/issues/25

### DIFF
--- a/lib/honeycomb/integrations/faraday.rb
+++ b/lib/honeycomb/integrations/faraday.rb
@@ -11,7 +11,7 @@ module Honeycomb
     end
 
     def call(env)
-      return if @client.nil?
+      return @app.call(env) if @client.nil?
 
       @client.start_span(name: "http_client") do |span|
         span.add_field "request.method", env.method.upcase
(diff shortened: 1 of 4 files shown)


Sergio-Mira:
(Looks like travis is missing the right ruby version `Required ruby-2.6.3 is not installed.`)

martin308:
@Sergio-Mira @DanielHeath this should fix the problem here, and I'll get this merged. 

I would probably advise setting the client to be a `NullClient` though for your tests as otherwise you will not be exercising a whole heap of code paths.

```
  config.client = Libhoney:: NullClient.new
end
```

martin308:
@Sergio-Mira added a test and a fix so that we continue to call any additional middleware that has been setup

lib/honeycomb/integrations/faraday.rb
```diff
@@ -11,6 +11,8 @@ def initialize(app, client:)
     end
 
     def call(env)
+      return @app.call(env) if @client.nil?
```
Sergio-Mira: 👍

PR #28, opened 2019-09-03
Update to release version of rails 6

martin308:
Rails 6 has been released now so we should test against that instead of rc2


martin308:
The build failures here should be fixed by https://github.com/honeycombio/libhoney-rb/pull/45 once it has been released

martin308:
To fix this we are going to "soft" stop supporting ruby 2.2, I've removed testing of 2.2 and updated various docs but it will technically work on 2.2 if people wrangle the dependencies themselves

Issue #31, opened 2019-10-04
ActionDispatch::Http::Parameters::ParseError with Rack app mounted in Rails


qnm:
I use https://github.com/twitchtv/twirp-ruby as part of a Rails app to provide Twirp RPC support.

I mount the Twirp app inside `config/routes.rb` as a rack app using mount e.g. `mount connections, :at => connections.full_name` where connections is a Twirp handler.

The issue I face is that the Rails instrumentation is applied during Twirp/rack requests, and the call to `request.param` in https://github.com/honeycombio/beeline-ruby/blob/master/lib/honeycomb/integrations/rails.rb#L16 returns the error `ActionDispatch::Http::Parameters::ParseError`.

One option may be to mount the Twirp app inside `config.ru` however mounting rack apps in Rails is a supported operation https://apidock.com/rails/ActionDispatch/Routing/Mapper/Base/mount

Thanks in advance for your help.

martin308:
@qnm thanks for the report! I'll take a look and see if I can reproduce and fix this issue. I think you're correct in that you should be able to mount racks apps in the standard way and have the beeline support this

martin308:
@qnm I've managed to reproduce this! Looking into a fix now

qnm:
@martin308 Only way I could think of was to supress the exception, which is a pretty blunt mechanism :(

martin308:
@qnm I've found a fix! Just trying to write a test to reproduce it and I'll push it up

martin308:
@qnm I've pushed up a fix for this [here](https://github.com/honeycombio/beeline-ruby/pull/39). I wasn't able to reproduce it in a test though. I tried it with a basic twirp handler which reproduced the issue and this fix resolved it. Let me know what you think

martin308:
@qnm this has now been released in [v1.1.1](https://github.com/honeycombio/beeline-ruby/releases/tag/v1.1.1) of the beeline

qnm:
Thanks @martin308 I hope to be able to test it today.

PR #39, opened 2019-10-09
Skip params when unavailable

martin308:
Fixes https://github.com/honeycombio/beeline-ruby/issues/31.

Tracing this through, we call this [method](https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/http/parameters.rb#L50) and `action_dispatch.request.parameters` is nil. So we fall into this [method](https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/http/request.rb#L381) which eventually ends up calling this [lambda](https://github.com/rails/rails/blob/a3d54d2afe557b243d6b18f9adf119f60ffb8242/actionpack/lib/action_dispatch/http/parameters.rb#L11) which raises an error when `raw_post` is nil. 

I'm not entirely clear what [twirp](https://github.com/twitchtv/twirp-ruby) is doing to the request to end up in this state and have struggled to reproduce this in a test. This should prevent the issue though and doesn't seem to cause any other side effects that I could find.

### DIFF
--- a/lib/honeycomb/integrations/rails.rb
+++ b/lib/honeycomb/integrations/rails.rb
@@ -13,6 +13,9 @@ module Honeycomb
 
       ::ActionDispatch::Request.new(env).tap do |request|
         # calling request.params will blow up if raw_post is nil
+        # the only known cause of this is when using the
+        # [twirp](https://github.com/twitchtv/twirp-ruby) rack app mounted in
+        # the rails app
         if request.raw_post
           yield "request.controller", request.params[:controller]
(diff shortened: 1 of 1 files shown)


martin308:
The circle ci build error will be gone once #38 is merged

toshok:
yeah, lgtm, although without more comments inline I'd have to git blame and grovel around in history to find this conversation.

martin308:
Added additional information to the comment, I'll wait until #38 has been merged though so that I can test that it can push to ruby gems correctly when I release this

PR #40, opened 2019-11-01
aws-sdk v2 & v3 integration

ajvondrak:
screenshot

This might technically be compatible with aws-sdk v2 as well, since v3 is supposed to be [API-compatible](https://github.com/aws/aws-sdk-ruby/blob/master/V3_UPGRADING_GUIDE.md). But I'm also not excited at supporting old software. :sweat_smile: Amazon is still apparently [supporting v2 of the gem](https://github.com/aws/aws-sdk-ruby/commits/version-2), but it seems to largely be automated maintenance (when they update the files that define the APIs/docs).

This started out by largely cribbing from the implementations of [`Aws::Plugins::ClientMetricsPlugin`](https://github.com/aws/aws-sdk-ruby/blob/d0c5f6e5a3e83eeda2d1c81f5dd80e5ac562a6dc/gems/aws-sdk-core/lib/aws-sdk-core/plugins/client_metrics_plugin.rb) and [`Aws::Plugins::ClientSendMetricsPlugin`](https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-core/lib/aws-sdk-core/plugins/client_metrics_send_plugin.rb). At first I thought that maybe we could just configure client metrics with a custom [publisher](https://github.com/aws/aws-sdk-ruby/blob/d0c5f6e5a3e83eeda2d1c81f5dd80e5ac562a6dc/gems/aws-sdk-core/lib/aws-sdk-core/client_side_monitoring/publisher.rb) that could feed spans to Honeycomb, but decided against that:

- The client metrics publisher batches and sends events all at once. We wouldn't easily be able to set the libhoney event timestamps + durations after-the-fact without exposing the guts of the `Honeycomb::Span`. I figured I didn't want to change the beeline internals that much.

- Hooking into the client metrics plugin means our available data would be dictated by the data in [`Aws::ClientSideMonitoring::RequestMetrics`](https://github.com/aws/aws-sdk-ruby/blob/d0c5f6e5a3e83eeda2d1c81f5dd80e5ac562a6dc/gems/aws-sdk-core/lib/aws-sdk-core/client_side_monitoring/request_metrics.rb). As I started building out the Honeycomb spans, it became clearer to me that I'd want more control over the fields being added. E.g., instead of sending a write-time synthesized field for [`final_error_retryable`](https://github.com/aws/aws-sdk-ruby/blob/d0c5f6e5a3e83eeda2d1c81f5dd80e5ac562a6dc/gems/aws-sdk-core/lib/aws-sdk-core/plugins/client_metrics_plugin.rb#L144-L150), we're free to decorate the Honeycomb span with raw configuration values for number of retries, the retry limit, and the final error type.

- Hand-rolling a new plugin + handlers just gives us more control to add a lot of rich metadata. Notably, the aws-api error handling is better than in [`Aws::Plugins::ClientMetricsSendPlugin`](https://github.com/aws/aws-sdk-ruby/blob/d0c5f6e5a3e83eeda2d1c81f5dd80e5ac562a6dc/gems/aws-sdk-core/lib/aws-sdk-core/plugins/client_metrics_send_plugin.rb#L63-L68), since I went through the effort of parsing response errors. But even just adding the `aws.params` field is already way more useful for diagnosing slow requests. The user can always configure a `presend_hook` to scrub fields they don't want. And if we want to add more information, we don't have to open aws-sdk-ruby pull requests or monkey-patch the CSM implementation.

The only big question marks in my mind are:

- Naming of the fields. I can't find a whole lot of consistency between different beelines, but I'm still not sure if the names I picked are running afoul of any strong conventions or anything.

- The development dependency was easy to test locally, but the appraisals + circleci configuration were added mostly blindly to parrot the other setups. Not sure if they'll work or if I missed anything.


findchris:
We are looking to use `aws-sdk v3` exclusively going forward, and this seems to check all of the boxes.  @martin308 - Can you have a look?

isnotajoke:
This looks really cool!

So, let's say I have a specific SDK call in my codebase, and I know that one of the parameters of that call is sensitive and shouldn't be sent outside of the application (maybe it's a password or a credit card number or something like that). It looks like this helpfully [puts parameters](https://github.com/honeycombio/beeline-ruby/pull/40/files#diff-2b0b8a59eea32634ec7a007d9efe9d0eR98) and some lower-level HTTP details on the span. It _seems_ like there's a potential for sensitive parameters to be sent by that logic, but I might be missing something. If that actually is a legitimate concern, is there to exclude certain operations from this reporting (at a specific callsite, for a particular parameter name, etc)?

ajvondrak:
> According to [this](https://ui.honeycomb.io/ruby-together/datasets/rubygems.org/result/JJoNuzPRYy5) the majority of downloads are still of v2.

Heyyy, how's _that_ for dog food? So cool! 😀

martin308:
@isnotajoke any sensitive parameters added to your spans can be scrubbed using a [`present_hook`](https://docs.honeycomb.io/getting-data-in/ruby/beeline/#augmenting-or-scrubbing-spans). Does that look like it will resolve your concerns?

@ajvondrak it's a really cool dataset! I've used it a bunch to make decisions about ruby versions and other library versions to support 😄

ajvondrak:
OK, added the tests for v2. Broke some of the tests, but they were pretty easy to fix, so I went ahead and patched them. Now we support v2 and v3! :tada:

Issue #41, opened 2019-11-07
No instrumentation from AWS integration when body consists of binary data


til:
When saving binary data, e.g. the result of zlib compressing a string, to an S3 object using aws-sdk-s3, there seems to be no corresponding instrumentation data received by Honeycomb for the related aws API calls.

When configuring Honeycomb locally with `config.client = Libhoney::LogClient.new(verbose: true)`, it outputs this error:

```
irb(main):002:0> bucket = Aws::S3::Resource.new(client: Aws::S3::Client.new(region: 'eu-central-1')).bucket('[redacted]')
irb(main):004:0> bucket.object('test.txt').put(body: Zlib.deflate('test'))
Honeycomb dataset '' | 2019-11-07T15:14:47+00:00 | Traceback (most recent call last):
        4: from (irb):4
        3: from libhoney (1.14.0) lib/libhoney/event.rb:126:in `send_presampled'
        2: from libhoney (1.14.0) lib/libhoney/client.rb:235:in `send_event'
        1: from libhoney (1.14.0) lib/libhoney/log_transmission.rb:25:in `add'
Encoding::UndefinedConversionError ("\x9C" from ASCII-8BIT to UTF-8)
```

A `pp` on the event data reveals this:

```
   "aws.params"=>
    {:body=>"x\x9C+I-.\x01\x00\x04]\x01\xC1",
     :bucket=>"[redacted]",
     :key=>"test.txt"},
```

So it looks like this is caused by calling `to_json` on `aws.params` which contains a string with a binary encoding.

I locally tried to remove the `:body` key from the `aws.params` hash before sending it, and the error went away.

If it helps – I wouldn't mind having the body data removed from instrumentation, especially as it can become quite large in the case of putting data on S3.

martin308:
@til thanks for the report! I'm taking a look now

til:
After upgrading libhoney to 1.14.1 which includes https://github.com/honeycombio/libhoney-rb/pull/37, we could see spans for `aws` calls being recorded on Honeycomb, so the most important part of this issue can be considered fixed now. (Side note: after confirming it works as intended, we ended up disabling the aws integration because we're uploading relatively large objects to s3, and were not sure if sending the same data again to Honeycomb would incur significant overhead).

When running it with `Libhoney::LogClient`, the error as initially reported still occurs. Do you plan to apply the same cleaning to the `Libhoney::LogClient` as well?

martin308:
[...]

PR #42, opened 2019-11-13
Redis integration

ajvondrak:
This implementation monkey-patches the `Redis::Client#process` method
that powers all the other `Redis::Client` methods who make requests to
the Redis server:

- `call`
- `call_loop`
- `call_pipeline`
- `call_pipelined`
- `call_with_timeout`
- `call_without_timeout`

Furthermore, `Redis::SubscribedClient`, `Redis::Distributed`, and
`Redis::Cluster` all wind up using instances of `Redis::Client`
underneath, so all those bases are covered with one relatively
straightforward patch.

I don't think the 2.x line of redis-rb is necessary to add. For one,
it's from 2011, and I think people have had plenty of time to upgrade.
The rubygems.org data bears this out [1]. 2.x also still supports Ruby
1.8 [2], which is well below the beeline's minimum supported Ruby
version.

The hard part about this integration is just the need for extensive
testing. The Redis API is huge, all manner of versions in the 3.x and
4.x lines are still in use, and the redis-rb maintainers have the habit
of adding/modifying commands across patch versions.

For the sake of completeness/paranoia, the tests in this commit have
been run successfully on *every* version of redis-rb between 3.0.0 and
4.1.3 on Ruby 2.6.3 (so that's: 3.0.0, 3.0.1, 3.0.2, 3.0.3, 3.0.4,
3.0.5, 3.0.6, 3.0.7, 3.1.0, 3.2.0, 3.2.1, 3.2.2, 3.3.0, 3.3.1, 3.3.2,
3.3.3, 3.3.5 [note that 3.3.4 was yanked], 4.0.0, 4.0.1, 4.0.2, 4.0.3,
4.1.0, 4.1.1, 4.1.2, and 4.1.3). Because the list is so extensive,
though, I didn't think it was wholly sane to iterate through the
explosion of combinations possible for CircleCI (every Redis version on
every Ruby version...). So I kept the Appraisals simply at the latest
redis 3.x + 4.x gems.

References:

[1] https://ui.honeycomb.io/ruby-together/datasets/rubygems.org/result/J4S6eMgLbv3

[2] https://github.com/redis/redis-rb/tree/be3d7e0b4194fcb7e91bee9153daf6073dbc1268#testing


martin308:
@findchris that's the plan. I'm working on fixing #41 first, then I'll get to reviewing this

lib/honeycomb/integrations/redis.rb
```diff
@@ -0,0 +1,221 @@
...
+        elsif char =~ /[[:print:]&&[:ascii:]]/
+          char
+        else
+          bytes = char.unpack("H*").pack("H*").bytes
+          bytes.map { |b| Kernel.format("\\x%02x", b) }.join
+        end
+      end
+
+      def needs_quotes?(char)
+        char =~ /[^[:print:]&&[:ascii:]]|[\\"\n\r\t\a\b' ]/
```
martin308: can we get a comment on this regex?

lib/honeycomb/integrations/redis.rb
```diff
@@ -0,0 +1,221 @@
...
+      # @see https://github.com/antirez/redis/blob/0f026af185e918a9773148f6ceaa1b084662be88/src/sds.c#L940-L1067
+      # @see https://github.com/antirez/redis/blob/0f026af185e918a9773148f6ceaa1b084662be88/src/sds.c#L878-L907
+      def escape(char)
+        if ESCAPES.key?(char)
+          ESCAPES[char]
+        elsif char =~ /[[:print:]&&[:ascii:]]/
+          char
+        else
+          bytes = char.unpack("H*").pack("H*").bytes
+          bytes.map { |b| Kernel.format("\\x%02x", b) }.join
```
martin308: can we get a comment on what this is doing?

lib/honeycomb/integrations/redis.rb
```diff
@@ -0,0 +1,221 @@
...
+      }.freeze
+
+      # This aims to replicate the algorithm used by redis-cli.
+      #
+      # @see https://github.com/antirez/redis/blob/0f026af185e918a9773148f6ceaa1b084662be88/src/sds.c#L940-L1067
+      # @see https://github.com/antirez/redis/blob/0f026af185e918a9773148f6ceaa1b084662be88/src/sds.c#L878-L907
+      def escape(char)
+        if ESCAPES.key?(char)
+          ESCAPES[char]
+        elsif char =~ /[[:print:]&&[:ascii:]]/
```
martin308: can we get a comment on this regex?

spec/honeycomb/integrations/redis_spec.rb
```diff
@@ -0,0 +1,2192 @@
...
+
+          it "sends just one event" do
+            expect(libhoney_client.events.size).to eq 1
+          end
+
+          it "sends the expected fields" do
+            expect(event_data.first).to include(
+              "meta.span_type" => "root",
+              "redis.command" => "FLUSHDB",
+              "redis.lo
```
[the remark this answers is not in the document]
martin308: I'm not sure if that suggestion will commit, looks like it's having trouble with line numbers but hopefully that's enough to reproduce it at least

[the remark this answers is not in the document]
ajvondrak: Good find! 👍

ajvondrak:
@martin308 Thanks for the review! I maybe went a tad overboard the refactoring/commentary, but it should at least be easier to understand now. Let me know what you think. 😄 

Also gave me a chance to realize that the `unpack` + `pack` combo from before was quite unnecessary, since I was just calling `String#bytes` anyway. 😅

PR #43, opened 2019-11-13
Fix for AWS body consisting of binary data

martin308:
Fix for #41. If we see that a string is being added as part of the `aws.params` then we check that it is UTF-8 encoded and substitute it if it is not.


lib/honeycomb/integrations/aws.rb
```diff
@@ -106,8 +106,16 @@ def setup(context)
           "aws.region" => context.config.region,
           "aws.service" => context.client.class.identifier,
           "aws.operation" => context.operation_name,
-          "aws.params" => context.params,
         }
+
+        context.params.each do |key, value|
+          if value.respond_to?(:encoding) && value.encoding != Encoding::UTF_8
+            value = "[#{value.encoding} encoded string]"
```
maplebed: Am I reading this right that if it is not UTF-8, it uses the string "foo encoded string" as the content and discards the actual value? I'm curious - are there any other text-encodings (ASCII? UTF-16?) that would be worth allowing? 

Are there any circumstances in which the value would not `respond_to(:encoding)`? If it doesn't respond, should the value be used or scrubbed?

martin308: That is correct, I'll see if I can handle this a different way as it is not actually the strings encoding that breaks the JSON encoding it is invalid characters in the string

martin308: The real fix I think is to resurrect https://github.com/honeycombio/libhoney-rb/pull/37 and get it working properly

martin308:
Closing in favour of https://github.com/honeycombio/libhoney-rb/pull/37

PR #44, opened 2019-11-19
Separate AWS context params

martin308:
I noticed while looking into #41 that the params are being added as an object when they could be split out into separate params. @ajvondrak and others using the AWS integration, does this look like it would be more useful for you?


ajvondrak:
This seems like a good idea to me! 👍

I suppose there's an option in honeycomb to expand nested JSON instead separate fields, but wouldn't want to force people to enable that on _everything_ just to get to the AWS params.

ajvondrak:
Question: would this be considered a breaking change, since folks might have queries built against `aws.params` already? 🤔 Just curious what the versioning policy would be on changing the fields that get sent by the beeline. Seems like such a little change to trigger a major version bump, lol.

martin308:
Yeh this would be a breaking change due to the field name changes. I'm tempted to say that this integration hasn't been out too long and maybe this is a small enough change that we could sneak this out in a non major version bump but it feels bad™️. I'll think a bit more on it and see what others internally think

grncdr:
I also think this would be worth it. (There's a similar issue with the `process_action.action_controller.params` field).

PR #46, opened 2019-11-26
Expose context methods for current trace and span

polotek:
Expose the current trace and current span in the top level scope. These properties are useful for doing custom instrumentation.

### DIFF
--- a/lib/honeycomb-beeline.rb
+++ b/lib/honeycomb-beeline.rb
@@ -25,7 +25,8 @@ module Honeycomb
     extend Forwardable
     attr_reader :client
 
-    def_delegators :@client, :start_span, :add_field, :add_field_to_trace
+    def_delegators :@client, :start_span, :add_field, :add_field_to_trace,
+                   :current_span, :current_trace
 
     def configure
       Configuration.new.tap do |config|
--- a/lib/honeycomb/client.rb
+++ b/lib/honeycomb/client.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "forwardable"
 require "honeycomb/beeline/version"
 require "honeycomb/configuration"
 require "honeycomb/context"
@@ -7,6 +8,10 @@ require "honeycomb/context"
 module Honeycomb
   # The Honeycomb Beeline client
   class Client
+    extend Forwardable
+
+    def_delegators :@context, :current_span, :current_trace
+
     def initialize(configuration:)
       @client = configuration.client
       # attempt to set the user_agent_addition, this will only work if the
--- a/lib/honeycomb/client.rb
+++ b/lib/honeycomb/client.rb
@@ -11,7 +11,6 @@ module Honeycomb
     extend Forwardable
 
     def_delegators :@context, :current_span, :current_trace
-    def_delegators :@client, :dataset, :service_name
 
     def initialize(configuration:)
       @client = configuration.client
@@ -80,10 +79,6 @@ module Honeycomb
       return if context.current_span.nil?
 
       context.current_span.trace.add_field("app.#{key}", value)
-    end
-
-    def debug?
-      client.debug ? true : false
     end
 
     private
--- a/lib/honeycomb/client.rb
+++ b/lib/honeycomb/client.rb
@@ -11,6 +11,7 @@ module Honeycomb
     extend Forwardable
 
     def_delegators :@context, :current_span, :current_trace
+    def_delegators :@client, :dataset, :service_name
 
     def initialize(configuration:)
       @client = configuration.client
@@ -79,6 +80,10 @@ module Honeycomb
       return if context.current_span.nil?
 
       context.current_span.trace.add_field("app.#{key}", value)
+    end
+
+    def debug?
+      client.debug ? true : false
     end
 
     private


lib/honeycomb/client.rb
```diff
@@ -1,12 +1,18 @@
...
 require "honeycomb/configuration"
 require "honeycomb/context"
 
 module Honeycomb
   # The Honeycomb Beeline client
   class Client
+    extend Forwardable
+
+    def_delegators :@context, :current_span, :current_trace
+    def_delegators :@client, :dataset, :service_name
```
ajvondrak: While we can add these couple of delegators to the `Libhoney::Client` object, I'd kind of like to get access to just the plain old libhoney client itself. E.g., I initially had something like this in my initializers:
```ruby
  config.client = Libhoney::TestClient.new if Rails.env.test?
end
```
But then using `TestClient` is pointless because I can't get access to the event data from `Honeycomb.client` at all. I.e., I can only get to the `Honeycomb::Client`, not the `Libhoney::Client`. So instead I have to do something in the test helper like
```ruby
$libhoney = Libhoney::TestClient.new
  config.client = $libhoney
end
```
Accessing the libhoney client from the `Honeycomb::Client` is awkward because of the naming, though. `Honeycomb.client.client` is...not ideal. 😅

As long as we're exposing new accessors, I'd be in favor of maybe mass-renaming the `Honeycomb::Client#client` to `Honeycomb::Client#libhoney`. The name would make the code in here a bit less confusing, IMO. And then public access would be fine:

```ruby
Honeycomb.client.libhoney.dataset
Honeycomb.client.libhoney.service_name
Honeycomb.client.libhoney.events # in the case of Libhoney::TestClient
Honeycomb.client.libhoney.add(data) # add global data to the underlying Libhoney::Builder
```

lib/honeycomb/client.rb
```diff
@@ -1,12 +1,18 @@
...
 require "honeycomb/beeline/version"
 require "honeycomb/configuration"
 require "honeycomb/context"
 
 module Honeycomb
   # The Honeycomb Beeline client
   class Client
+    extend Forwardable
+
+    def_delegators :@context, :current_span, :current_trace
```
ajvondrak: Noting here that it might be worth using these delegators across the `Honeycomb::Client` code instead of `context.current_span` all over the place. 💅 ✨

lib/honeycomb-beeline.rb
```diff
@@ -25,7 +25,8 @@ class << self
     extend Forwardable
     attr_reader :client
 
-    def_delegators :@client, :start_span, :add_field, :add_field_to_trace
+    def_delegators :@client, :start_span, :add_field, :add_field_to_trace,
+                   :current_span, :current_trace
```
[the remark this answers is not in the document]
polotek: I agree that the ideal design is one where honeycomb instrumentation should be as transparent as possible. However, I'm wary of introducing too much magic or cruft around these things.

My preference would be have the top level `add` methods do nil checking. But leave the `current` methods as is. So in most instrumentation you can just do `Honeycomb.add_field('error', err)` and be assured it won't break anything, even if there's no trace. But `current_trace` and `current_span` are named explicitly so you know what you're getting. And having them return `nil` is also a relevant indicator.

I'm considering adding an `active_trace?` checker as well. That appears in other beeline libraries. I'm gonna wait for input from the HC folks first though.

lib/honeycomb/client.rb
polotek: Keep in mind that this PR does not expose the libhoney client. Part of my goal here is to have more of a conversation with HC about establishing some thoughtful standards here. I can think of reasons to prefer hiding the libhoney client vs exposing it. In any app you are configuring, you can always keep your own reference to the client yourself. But giving third-party code the ability to manipulate the underlying client directly may not be the right thing.

grncdr:
Just want to add that since upgrading from 0.8, I couldn't find any "official" way of creating a tracing header for propagation, and ended up with this code which is using `send` to hack around private methods:

```rb
trace_header = Honeycomb.client.send(:context).current_span.try(&:to_trace_header)
```

So any PR that let's me avoid that would be most welcome 👍

lib/honeycomb/client.rb
martin308: I think the idea of exposing current span and current trace are good and I can see the need for them. What were you thinking would be the use case for exposing these (`dataset` and `service_name`)? I'd like to strike a good balance between usability and limiting the public API so that we can make internal changes easier

lib/honeycomb/client.rb
```diff
@@ -76,6 +82,10 @@ def add_field_to_trace(key, value)
       context.current_span.trace.add_field("app.#{key}", value)
     end
 
+    def debug?
+      client.debug ? true : false
+    end
+
```
martin308: same question here as above, what is the use case that you were seeing for exposing this?

martin308:
@grncdr the way I was expecting people to use the trace header stuff was that you would have a span wrapping the http request so you would end up with something like the faraday integration https://github.com/honeycombio/beeline-ruby/blob/master/lib/honeycomb/integrations/faraday.rb#L25.

lib/honeycomb/client.rb
ajvondrak: > I'd like to strike a good balance between usability and limiting the public API so that we can make internal changes easier

The options I see:

1. Expose the `Honeycomb::Client#client` publicly.

    **Pros**
    * Limits the `Honeycomb::Client` API to just the one attribute instead of a bunch of delegators.
    * Maximally flexible: users can set `Honeycomb::Configuration#client` to objects with more exotic APIs. We can already set it to whatever object we want via `Honeycomb.configure { |config| config.client = ... }`. Not then being able to access it is a big inconvenience. E.g., even just being able to set a `Libhoney::TestClient` and get access to the `#events`, you have to `Honeycomb.configure` every time or keep your own global reference or whatnot.

    **Cons**
    * Bad name: `Honeycomb.client.client`.
    * Exposes the `@api private` guts of `Libhoney::Client`, like `#send_event`. (Although nothing's to stop that if people just drop the beeline and use libhoney directly anyway. :man_shrugging:)

2. Expose the public API of `Libhoney::Client` via delegators in `Honeycomb::Client`.

    **Pros**
    * Doesn't expose `@api private` guts of libhoney (the things that are technically public but considered private).
    * Even just exposing `#add`, `#add_field`, and `#add_dynamic_field` unlocks so much functionality to add global - possibly dynamic - data to your traces.
    * Attribute-wise, I think `writekey`, `dataset`, `sample_rate`, and `api_host` are easily the high-level things that people care about. The other attributes, not so much.

    **Cons**
    * Tightly couples the API of `Honeycomb::Client` to the API of `Libhoney::Client`. I.e., more methods wind up on `Honeycomb::Client`.
    * If you want to be maximally flexible, there's no picking & choosing. To expose one public API from libhoney, you'd better expose every other public API. Otherwise, it's the same pattern: "hey, they don't expose xyz, I have to drop down to libhoney". Today someone wants `dataset`, tomorrow...`api_host`? `block_on_send`? etc

3. Keep it as-is.
    **Pros**
    * No changes needed. :wink:
    * If the answer to any of the above is "just keep a global reference to your `Libhoney::Client` object", then why add delegators in the first place?

    **Cons**
    * Kind of a pain.
    * Could avoid the global variable (since they're generally frowned upon, obviously) with a public accessor.

4. Expose the current `Honeycomb::Configuration` (e.g., `Honeycomb.configuration`).
    **Pros**
    * Gives us access to the "guts" via (say) `Honeycomb::Configuration#client`.
    * Keeps the API essentially the same: we're already setting the attributes, why not let us get them after the fact?
    * Don't have to mass-rename `Honeycomb::Client#client`, since `Honeycomb.configuration.client` reads fine.

    **Cons**
    * I still personally think the difference between `Honeycomb.client` and `Honeycomb.configuration.client` is confusing. Calling two different things "honeycomb [something something] clients".
    * Exposes the attribute writers as well, so you could (for example) call `Honeycomb.configuration.presend_hook` outside of a `Honeycomb.configure` block. Could this be solved by `#freeze`-ing the config? 🤔

ajvondrak:
> Would love to have a little bit more discussion on exposing the non `current_span` and `current_trace` accessors. We could split that into another PR so we can discuss those features and get `current_span` and `current_trace` released?

Big :+1: on that.

lib/honeycomb-beeline.rb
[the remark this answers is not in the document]
ajvondrak: > I think I prefer having `current_span` and `current_trace` as they are more descriptive than just having `span` and `trace`.

The way I see it: what span/trace would I be accessing if not the current one? :man_shrugging: I run into my 80-character line limits fast enough with field names, I'd like the help of not having a redundant `current_`, lol.

Either way, throwing in `alias_method` gives people the option; pretty common Ruby-ism. But whatever, not a hill I'm gonna die on.

> I also agree that having these return nil would be more intuitive.

I was thinking of it like an expansion to get rid of `Honeycomb.add*`, like: `Honeycomb.add_field` => `Honeycomb.current_span.add_field`. So the null-object pattern makes the API more consistent.

However, the secondary purpose of `Honeycomb.add_field` (other than nil-safety) is to add the `app.` namespace. Now that I understand the context of using `app.` from the Slack discussion, I think the namespacing is important to keep intact. Given the `app.`-ening is extra functionality that `Honeycomb.current_span.add_field` would **not** emulate, I agree that nil-safety also need not be emulated.

TL;DR: straight delegator works for me. 😀

lib/honeycomb/client.rb
martin308: I think it's hard for me to evaluate these options without knowing the use case for accessing the `dataset` and `service_name` property that the beeline is using. Could you elaborate on what problem these various options are solving?

ajvondrak: I'm not sure about the `dataset`/`service_name`/`debug` being exposed in this PR. @polotek?

For my part, I care more about:
* `Libhoney::TestClient#events`, per my initial comment about being able to put the initializer logic one place.
* `Libhoney::Client#{add,add_field,add_dynamic_field}`. For instance, I want to add my application's version to the whole trace by default, so calling `#add_field` on the global builder is the most direct way of doing that. Right now I do that with
    ```ruby
      config.dataset = 'app'

      libhoney = App.env.test? ? Libhoney::TestClient : Libhoney::Client
      libhoney = libhoney.new(writekey: config.write_key, dataset: config.dataset)
      libhoney.add_field('app.version', App.version)

      config.client = libhoney
    end
    ```
    :point_up: Note that I can't simply say
    ```ruby
    config.client = Libhoney::TestClient.new if App.env.test?
    config.client.add_field('app.version', App.version)
    ```
    because `Honeycomb::Configuration#client` is not memoized, so the default value that gets constructed will be *reconstructed* when the `Honeycomb.configure` block returns and we ultimately set `Honeycomb.client = Client.new(configuration: ...)`. So the field added to the builder gets scrapped! But I figured the first version was clearer than
    ```ruby
    config.client = Libhoney::TestClient.new if App.env.test?
    config.client = config.client # save the default value when not in test env
    config.client.add_field('app.version', App.version)
    ```

polotek: The primary use case for exposing dataset is so you can create trace urls. https://docs.honeycomb.io/api/direct-trace-links/

You need these values:
- `team` - not part of the beeline context. The caller should have this value on their side.
- `dataset` - the caller could also keep this context on their side. But it's already configured in the beeline, just not exposed. And it's possible to have multiple datasets in place in the same system by configuring multiple clients.
- `trace_id` - this is the primary reason to get the `current_trace`.

Why are we creating trace urls? We also have error reporting going to bugsnag in our system. We have started to create trace urls and send them in the bugsnag context. So when we look up error information, we can click straight through to the honeycomb trace. It is fantastic and I want to do that in more places.

polotek: My primary concern in this PR is enabling trace urls. However I also exposed the other values and you asked about that. It's a fair question.

My thinking is that it generally makes sense to expose the configuration that was used to create the libhoney client. It's read-only, and we are finding it is sometimes useful to inspect the context of the environment to make decisions.

That said, I'm happy to break those out into a separate PR. It'll probably be the end of next week before I get to this.

polotek: Per conversation and feedback, I removed the configuration stuff. That can go to a separate PR.

lib/honeycomb/client.rb
polotek: Removed.

polotek:
This PR is super simple now. Just exposing the `current_trace` and `current_span`. I think there's a lot more to discuss. But if we can land this, it'll enable a lot of good things that aren't possible today.

PR #50, opened 2019-12-11
Handle bad request params

martin308:
As mentioned in #49, the rails middleware can break when there are bad params sent to the rails application and it tries to read out the controller and action. 

In this we catch the bad request exception and set the message as the value for controller and action. I'm not sure if this super useful though. A sample of this is below. Anyone have any thoughts? We could also just not add those keys at all or set them to 'unknown' or similar?

```
"request.action"=>"Invalid query parameters: Invalid encoding for parameter: �"
```


cade:
> Anyone have any thoughts?

I don't have any preferences on what's recorded, but I'll check with the Ops team here to see if they've got any feedback!

cade:
> I'll check with the Ops team

They didn't have any suggestions!

PR #51, opened 2019-12-12
Add Honeycomb.{with_field,with_trace_field}

ajvondrak:
The return value of `Honeycomb.add_field` (and similarly
`Honeycomb.add_field_to_trace`) is the `Libhoney::Event` object that had
the field added.

However, I'd often have the pattern in my code of using

```ruby
def something
  (some || sort && of ** expression).tap do |value|
    Honeycomb.add_field('field', value)
  end
end
```

That is, I still want the return value of the `#something` method, but I
wanted to track its value in a Honeycomb field. The methods added in
this commit allow the more natural pattern of wrapping the computation
in a block like so:

```ruby
def something
  Honeycomb.with_field('field') do
    some || sort && of ** expression
  end
end
```

It's just a minor feature, but one that would make some of my code a
little nicer looking. Especially with predicates, where I currently have
code like `question?.tap { ... }`.

### DIFF
--- a/lib/honeycomb-beeline.rb
+++ b/lib/honeycomb-beeline.rb
@@ -25,7 +25,12 @@ module Honeycomb
     extend Forwardable
     attr_reader :client
 
-    def_delegators :@client, :start_span, :add_field, :add_field_to_trace
+    def_delegators :@client,
+                   :start_span,
+                   :add_field,
+                   :add_field_to_trace,
+                   :with_field,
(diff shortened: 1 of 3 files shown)


MikeGoldsmith:
@ajvondrak Thank you for the contribution and sorry for the _super_ slow response. I've merged main into your PR and resolved conflicts so we're all good to accept it. Thanks you!

PR #60, opened 2020-01-09
Fix `Redis.honeycomb_client = nil`

ajvondrak:
The idea of this attribute was that you'd be able to turn off the Redis
instrumentation by blanking out the Honeycomb client.

However, by using `||`, the code was *actually* falling back to the
global `Honeycomb.client` whenever we set `Redis.honeycomb_client =
nil`. The spec was passing erroneously because it was being run in a
context where `Honeycomb.client` was *also* `nil`. Oops!

By switching this to use the `defined?` check, we'll return `nil` from
`Redis.honeycomb_client` as long as it's been explicitly blanked out by
the user.

### DIFF
--- a/lib/honeycomb/integrations/redis.rb
+++ b/lib/honeycomb/integrations/redis.rb
@@ -23,7 +23,9 @@ module Honeycomb
       attr_writer :honeycomb_client
 
       def honeycomb_client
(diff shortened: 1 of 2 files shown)


ajvondrak:
Looks like the only failures in CirlceCI are from the Faraday issue fixed by #59.

Issue #62, opened 2020-01-13
Rack Integration Raises an Error During File Upload


mstruve:
I recently noticed that we have had a few errors raised in our production environment from the beeline gem when a user tries to upload a file. They seem pretty deep in the rack implementation so I figured I would share here to see if you all had any suggestions for a fix. 

Here is the error and backtrace:
```
NoMethodError: undefined method `empty?' for #<ActionDispatch::Http::UploadedFile:0x00000000085b38a0>
 rack.rb  31 block (2 levels) in call(...)
[GEM_ROOT]/gems/honeycomb-beeline-1.3.0/lib/honeycomb/integrations/rack.rb:31:in `block (2 levels) in call'
29       client.start_span(name: "http_request", serialized_trace: hny) do |span|
30         add_field = lambda do |key, value|
31           next unless value && !value.empty?
32 
33           span.add_field(key, value)

 [GEM_ROOT]/gems/honeycomb-beeline-1.3.0/lib/honeycomb/integrations/rails.rb:14 :in `tap`
 [GEM_ROOT]/gems/honeycomb-beeline-1.3.0/lib/honeycomb/integrations/rails.rb:14 :in `add_package_information`
 [GEM_ROOT]/gems/honeycomb-beeline-1.3.0/lib/honeycomb/integrations/rack.rb:40 :in `block in call`
 [GEM_ROOT]/gems/honeycomb-beeline-1.3.0/lib/honeycomb/client.rb:54 :in `start_span`
 [GEM_ROOT]/gems/honeycomb-beeline-1.3.0/lib/honeycomb/integrations/rack.rb:29 :in `call`
 [GEM_ROOT]/gems/actionpack-5.2.4.1/lib/action_dispatch/middleware/remote_ip.rb:81 :in `call`
 [GEM_ROOT]/gems/request_store-1.4.1/lib/request_store/middleware.rb:19 :in `call`
 [GEM_ROOT]/gems/rack-timeout-0.6.0/lib/rack/timeout/core.rb:151 :in `block in call`
 [GEM_ROOT]/gems/rack-timeout-0.6.0/lib/rack/timeout/support/timeout.rb:19 :in `timeout`
 [GEM_ROOT]/gems/rack-timeout-0.6.0/lib/rack/timeout/core.rb:150 :in `call`
 [GEM_ROOT]/gems/actionpack-5.2.4.1/lib/action_dispatch/middleware/request_id.rb:27 :in `call`
[...]

PR #64, opened 2020-01-24
Avoid exception when checking uploaded files

DanielHeath:
Fixes https://github.com/honeycombio/beeline-ruby/issues/62


DanielHeath:
Ahh - this is an awkward place to do it, because you can't assume rails is available from the rack handler. Still, this branch solves *my* problem, so 'hooray'.

martin308:
@DanielHeath this should be resolved in #73 which will be released shortly as v2 of the beeline!

PR #65, opened 2020-01-24
Allow the app to handle invalid JSON

DanielHeath:
If a POST is sent with `content-type: application/json` but the body is invalid JSON, beeline-ruby blows up.

In my app, this indicates a third-party developer misusing the API, and we respond with a page explaining what's happened to them. Introducing beeline-ruby broke that.


martin308:
@DanielHeath this should be resolved in #73 which will be released shortly as v2 of the beeline!

PR #69, opened 2020-02-05
Add fields found in other beelines.

paulosman:
Several other beelines include the count and names of instrumentations,
so we'll include them here. Also additional headers and metadata about
the HTTP request and response in rack middleware.


lib/honeycomb/integrations/rack.rb
```diff
@@ -13,8 +13,11 @@ module Rack
       ["HTTP_VERSION", "request.http_version"],
       ["HTTP_HOST", "request.host"],
       ["REMOTE_ADDR", "request.remote_addr"],
+      ["HTTP_ACCEPT", "request.header.accept"],
+      ["HTTP_CONTENT_TYPE", "request.header.content_type"],
       ["HTTP_USER_AGENT", "request.header.user_agent"],
       ["rack.url_scheme", "request.protocol"],
+      ["rack.url_scheme", "request.scheme"],
```
martin308: do the other beelines have both `request.protocol` and `request.scheme`? Should we remove one of these?

lib/honeycomb/integrations/rack.rb
```diff
@@ -33,15 +36,25 @@ def call(env)
...
+        is_secure = lambda do
+          env["rack.url_scheme"] == "https" || env["rack.url_scheme"] == "wss"
+        end
+
+        is_xhr = -> { env["HTTP_X_REQUESTED_WITH"] == "XMLHttpRequest" }
+
         extract_fields(env, RACK_FIELDS, &add_field)
 
+        span.add_field("request.secure", is_secure.call)
+        span.add_field("request.xhr", is_xhr.call)
```
martin308: why the procs/lambdas here over just variables?

lib/honeycomb/client.rb
```diff
@@ -22,6 +22,10 @@ def initialize(configuration:)
       @client.add_field "meta.beeline_version", Honeycomb::Beeline::VERSION
       @client.add_field "meta.local_hostname", configuration.host_name
 
+      integrations = Honeycomb.integrations_to_load
+      @client.add_field "meta.instrumentations_count", integrations.count
+      @client.add_field "meta.instrumentations", integrations.map(&:to_s).to_s
```
martin308: it would be good to add these fields to https://github.com/honeycombio/beeline-ruby/blob/master/spec/support/event_data_shared_examples.rb#L3 so that they are included in the base tests

lib/honeycomb/integrations/rack.rb
```diff
@@ -13,8 +13,11 @@ module Rack
       ["HTTP_VERSION", "request.http_version"],
       ["HTTP_HOST", "request.host"],
       ["REMOTE_ADDR", "request.remote_addr"],
+      ["HTTP_ACCEPT", "request.header.accept"],
+      ["HTTP_CONTENT_TYPE", "request.header.content_type"],
```
martin308: Can we add all of these new fields to https://github.com/honeycombio/beeline-ruby/blob/master/spec/support/event_data_shared_examples.rb#L17?

lib/honeycomb/integrations/rack.rb
paulosman: @martin308 I went back and forth on that :) The short version is that the lambdas make the code a bit shorter and in the case of `is_secure` at least, encapsulates some business logic (the environment value can be x or y).

lib/honeycomb/integrations/rack.rb
paulosman: Most have `scheme` but not `protocol`. I agree that both is less than ideal - I'll remove `protocol`.

lib/honeycomb/integrations/rack.rb
martin308: I don't think I have strong feelings either way, it just stood out as being a bit different. Is there any performance implication to doing it this way instead of just assigning them to variables?

paulosman: @martin308 I replaced with just straight forward assignment anyway. Sat with it and don't think it justifies the difference.

martin308: ok cool. I did also just look up the rack docs and potentially we could just rely on the rack methods for this and maybe some of the other bits in here. I don't remember if I considered that when adding these fields originally. There is [xhr?](https://www.rubydoc.info/gems/rack/Rack%2FRequest%2FHelpers:xhr%3F) and [ssl?](https://www.rubydoc.info/gems/rack/Rack%2FRequest%2FHelpers:ssl%3F)

lib/honeycomb/integrations/rack.rb
paulosman: Good catch, done!

PR #70, opened 2020-02-06
Make the libhoney client instance accessible.

paulosman:
There are cases where a beeline user would like to get access to the
libhoney instance directly. This makes it possible by calling
`Honeycomb.libhoney` at any time.

### DIFF
--- a/lib/honeycomb-beeline.rb
+++ b/lib/honeycomb-beeline.rb
@@ -25,8 +25,8 @@ module Honeycomb
     extend Forwardable
     attr_reader :client
 
-    def_delegators :@client, :start_span, :add_field, :add_field_to_trace,
-                   :current_span, :current_trace
+    def_delegators :@client, :libhoney, :start_span, :add_field,
+                   :add_field_to_trace, :current_span, :current_trace
 
     def configure
       Configuration.new.tap do |config|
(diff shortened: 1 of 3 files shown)


lib/honeycomb-beeline.rb
```diff
@@ -25,8 +25,8 @@ class << self
     extend Forwardable
     attr_reader :client
 
-    def_delegators :@client, :start_span, :add_field, :add_field_to_trace,
-                   :current_span, :current_trace
+    def_delegators :@client, :libhoney, :start_span, :add_field,
+                   :add_field_to_trace, :current_span, :current_trace
```
martin308: maybe we could add a basic test to this file for this new delegator? https://github.com/honeycombio/beeline-ruby/blob/master/spec/honeycomb/beeline_spec.rb

paulosman: Good idea - done!

spec/honeycomb/beeline_spec.rb
```diff
@@ -14,6 +14,12 @@
     end
   end
 
+  describe "accessing the libhoney instance" do
+    it "contains the expected field" do
+      expect(Honeycomb.libhoney).to eq libhoney_client
+    end
+  end
+
```
martin308: 🙌

PR #72, opened 2020-02-07
Add ruby 2.7 to the test matrix

martin308:
Adds ruby 2.7 to the test matrix now that it has been released. Didn't add rails 4.2 to the ruby 2.7 matrix as it would require some bundler version wrangling. Maybe it would be easy to add in, I didn't spend a lot of time investigating.


martin308:
Now that we have #87 this is redundant and 2.7 will need to be added to the matrix

PR #73, opened 2020-02-22
Rework the Rails integration

### LINKED ISSUE #31: ActionDispatch::Http::Parameters::ParseError with Rack app mounted in Rails

### LINKED ISSUE #62: Rack Integration Raises an Error During File Upload

ajvondrak:
This refactoring avoids ever parsing GET/POST parameters to obtain the
Rails routing information required for the `request.action`,
`request.controller`, and `request.route` fields.

This is accomplished in two ways:

1. Use `ActionDispatch::Request#path_parameters` instead of
   `ActionDispatch::Request#params`. The former is the subset of data we
   need for `request.action` + `request.controller`, and in fact are the
   values merged into the latter hash after `#params` parses the normal
   GET/POST parameters. So we can avoid the extraneous parsing by just
   getting the path parameters straight from the horse's mouth.

2. Call `ActionDispatch::Journey::Router#recognize` on a simplified
   request that *only* contains the necessary routing information. The
   `#recognize` method will yield the `ActionDispatch::Request#params`
   when it matches a route, which again triggers the GET/POST parsing.
   But we don't care about the HTTP parameters. So instead I copied the
   approach of `ActionDispatch::Routing::RouteSet#recognize_path`, which
   reconstructs a simplified request with just the information that it
   needs. (Unfortunately we can't use `#recognize_path` directly because
   instead of returning the route it returns the path parameters, which
   we can already get directly from `#path_parameters`.)

This accomplishes several things:

* Fixes #62, since we don't get a chance to erroneously fall back to
  GET/POST parameters in the event that the routing information is
  missing (due to requests to unrecognized routes).

* Fixes #31, since the bug ultimately comes from the twirp gem causing
  POST parameter parsing to fail. But if we never parse the HTTP
  parameters, no problem. This is a more direct workaround than #39, but
  ultimately is a problem that should be fixed upstream.

* Fixes #49, since the parameter encoding won't matter if we never try
  to parse them. Renders #50 unnecessary, since we won't have a chance
  to trigger the exceptions in question.

* Gives us more data in more situations. The new implementation of the
  `request.route` field now comes up non-nil in the event of errors
  because we use the `ActionDispatch::Request#original_fullpath` and not
  the `#path_info` that gets [rewritten by
  `ActionDispatch::ShowExceptions`](https://github.com/rails/rails/blob/758e4f8406e680a6cbf21b170749202c537a2576/actionpack/lib/action_dispatch/middleware/show_exceptions.rb#L49).

* Gives us more data in more Rails versions. With the addition of actual
  tests for the field values, it became clear that `request.route` was
  missing in Rails 4, simply because of an interface change introduced
  in Rails 5.

---

Revamp the Rails integration's tests

* Remove extraneous spans that were being generated due to the
  ActiveSupport::Notifications subscriptions. There's no need to
  actually configure these from the tests, since we already have the
  active_support_spec.rb, and it's not like the config in the
  rails_spec.rb even matched the one from the Rails generator or
  anything. These just get in the way of testing the Rails integration,
  since the custom fields we'd want to make assertions about only wind
  up on the root span anyway.

* Factor out assertions into a set of shared examples, and actually
  start making assertions about the `request.{action,controller,route}`
  fields that the integration adds (which weren't even being tested for
  previously 😱).

* Rework the "bad request" tests that reproduce #49. These tests will
  now break against the old implementation, since they assert that
  routing information is still present (whereas #50 wipes the fields out
  with error messages). There are ways to get the routing information
  without ever trying to parse the GET/POST parameters, so we can still
  save this use case. Furthermore, the only Rails 5+ specific changes
  are to whether or not Rails responds with an HTTP 400, so we needn't
  wipe out the entire `describe` block for Rails < 5.

* Add tests to reproduce the issue described in #62, where unrecognized
  routes may cause the old implementation to erroneously fall back to
  the GET/POST parameters for routing information.

* Add tests that reproduce #31. The bug is actually in the twirp gem,
  whose middleware fails to rewind the `rack.input` after reading it.
  I've verified these regression tests against the old version of the
  Rails integration that didn't check `if request.raw_post`, and they
  fail appropriately. I wanted to have this test in here to have
  confidence in changing the implementation of the Rails integration.
  The `raw_post` workaround of #39 is unnecessary if we actually never
  look at the GET/POST parameters in the first place (no parsing, no
  problem).

### DIFF
--- a/spec/honeycomb/integrations/rails_spec.rb
+++ b/spec/honeycomb/integrations/rails_spec.rb
@@ -14,10 +14,6 @@ if defined?(Honeycomb::Rails)
     let(:configuration) do
       Honeycomb::Configuration.new.tap do |config|
         config.client = libhoney_client
-        config.notification_events = %w[
-          render_template.action_view
-          process_action.action_controller
-        ].freeze
       end
     end
     let(:client) { Honeycomb::Client.new(configuration: configuration) }
@@ -43,51 +39,127 @@ if defined?(Honeycomb::Rails)
 
     class TestController < ActionController::Base
       def hello
-        render plain: "Hello World!"
+        render plain: "Hello #{params[:name]}!"
+      end
+    end
+
+    shared_examples_for "the rails integration" do
+      let(:event_data) { libhoney_client.events.map(&:data) }
+
+      it_behaves_like "event data"
+
+      it "sends the right number of events" do
+        expect(libhoney_client.events.size).to eq 1
+      end
+
+      let(:event) { event_data.first }
+
+      it "sends the right request.controller" do
+        expect(event["request.controller"]).to eq controller
+      end
+
+      it "sends the right request.action" do
+        expect(event["request.action"]).to eq action
+      end
+
+      it "sends the right request.route" do
+        expect(event["request.route"]).to eq route
       end
     end
 
     describe "a standard request" do
       before do
-        header("Http-Version", "HTTP/1.0")
-        header("User-Agent", "RackSpec")
-
-        get "/hello/martin"
+        get "/hello/world"
       end
 
       it "returns ok" do
         expect(last_response).to be_ok
       end
 
-      it "sends the right number of events" do
-        expect(libhoney_client.events.size).to eq 3
+      include_examples "the rails integration" do
+        let(:controller) { "test" }
+        let(:action) { "hello" }
+        let(:route) { "GET /hello/:name(.:format)" }
       end
-
-      let(:event_data) { libhoney_client.events.map(&:data) }
-
-      it_behaves_like "event data"
     end
 
-    if VERSION >= Gem::Version.new("5")
-      describe "a bad request" do
-        before do
-          header("Http-Version", "HTTP/1.0")
-          header("User-Agent", "RackSpec")
+    describe "a request with invalid parameter encoding" do
+      before do
+        get "/hello/world?via=%c1"
+      end
 
-          get "/hello/martin?via=%c1"
-        end
-
+      if VERSION >= Gem::Version.new("5")
         it "returns bad request" do
           expect(last_response).to be_bad_request
         end
-
-        it "sends the right number of events" do
-          expect(libhoney_client.events.size).to eq 1
+      else
+        it "returns ok" do
+          expect(last_response).to be_ok
         end
+      end
 
-        let(:event_data) { libhoney_client.events.map(&:data) }
+      include_examples "the rails integration" do
+        let(:controller) { "test" }
+        let(:action) { "hello" }
+        let(:route) { "GET /hello/:name(.:format)" }
+      end
+    end
 
-        it_behaves_like "event data"
+    describe "an unrecognized request" do
+      before do
+        get "/unrecognized", action: "action", controller: "controller"
+      end
+
+      it "returns not found" do
+        expect(last_response).to be_not_found
+      end
+
+      include_examples "the rails integration" do
+        let(:controller) { nil }
+        let(:action) { nil }
+        let(:route) { nil }
+      end
+    end
+
+    describe "twirp bug" do
+      before do
+        app.routes.draw do
+          mount TwirpBug, at: "/twirp", via: :post
+        end
+      end
+
+      # Emulates a bug in the twirp gem.
+      #
+      # The twirp middleware calls IO#read on the rack.input without rewinding.
+      # Later, when Rails tries to parse the request parameters by calling
+      # IO#read(n) with the content length, we get back nil, which blows up the
+      # default JSON MIME type handler.
+      #
+      # @see https://github.com/honeycombio/beeline-ruby/issues/31
+      # @see https://github.com/honeycombio/beeline-ruby/pull/39
+      # @see https://github.com/twitchtv/twirp-ruby/blob/c8520030b3e4eb584042b0a8db9ae606a3b6c6f4/lib/twirp/service.rb#L138
+      module TwirpBug
+        def self.call(env)
+          env["rack.input"].read # don't rewind
+          ::Rack::Response.new("OK").finish
+        end
+      end
+
+      before do
+        env "CONTENT_TYPE", "application/json"
+        env "CONTENT_LENGTH", 17
+        env "rack.input", StringIO.new('{"json":"object"}')
+        post "/twirp"
+      end
+
+      it "returns ok" do
+        expect(last_response).to be_ok
+      end
+
+      include_examples "the rails integration" do
+        let(:controller) { nil }
+        let(:action) { nil }
+        let(:route) { "POST /twirp" }
       end
     end
   end
--- a/spec/honeycomb/integrations/rails_spec.rb
+++ b/spec/honeycomb/integrations/rails_spec.rb
@@ -10,6 +10,19 @@ if defined?(Honeycomb::Rails)
     VERSION = Gem::Version.new(::Rails::VERSION::STRING)
     include Rack::Test::Methods
 
+    # These headers are required for the HTTP fields from the Rack integration.
+    #
+    # In order to use the shared examples for event data with HTTP fields
+    # included, these headers have to be present. Additionally, every request
+    # made in this suite has to include a query string, so we just tack on
+    # `?honey=bee` to everything.
+    before do
+      header "Http-Version", "HTTP/1.0"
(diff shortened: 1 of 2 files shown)


ajvondrak:
Oof. Looks like the CircleCI failure is an issue with the AWS integration because the test picks a random client object to instantiate, which blew up when it wasn't given a required parameter:
```ruby
aws = aws_clients.sample.new(honeycomb_client: client)
```
```
     ArgumentError:
       missing required option `:endpoint'
     # ./gemfiles/vendor/bundle/ruby/2.3.0/gems/aws-sdk-core-2.11.450/lib/seahorse/client/plugins/endpoint.rb:25:in `after_initialize'
     # ./gemfiles/vendor/bundle/ruby/2.3.0/gems/aws-sdk-core-2.11.450/lib/seahorse/client/base.rb:84:in `block in after_initialize'
     # ./gemfiles/vendor/bundle/ruby/2.3.0/gems/aws-sdk-core-2.11.450/lib/seahorse/client/base.rb:83:in `each'
     # ./gemfiles/vendor/bundle/ruby/2.3.0/gems/aws-sdk-core-2.11.450/lib/seahorse/client/base.rb:83:in `after_initialize'
     # ./gemfiles/vendor/bundle/ruby/2.3.0/gems/aws-sdk-core-2.11.450/lib/seahorse/client/base.rb:21:in `initialize'
     # ./gemfiles/vendor/bundle/ruby/2.3.0/gems/aws-sdk-core-2.11.450/lib/seahorse/client/base.rb:105:in `new'
     # ./spec/honeycomb/integrations/aws_spec.rb:76:in `block (3 levels) in <top (required)>'
```
I posit we just got unlucky. The tests should probably be more deterministic. However, I don't think this failure should be a blocker to merging this PR.

ajvondrak:
This also looks like it'll fix #65.

martin308:
@ajvondrak this looks great! I’ll be able to give this a proper review next week

spec/honeycomb/integrations/rails_spec.rb
```diff
@@ -43,51 +39,127 @@
...
       def hello
-        render plain: "Hello World!"
+        render plain: "Hello #{params[:name]}!"
+      end
+    end
+
+    shared_examples_for "the rails integration" do
+      let(:event_data) { libhoney_client.events.map(&:data) }
+
+      it_behaves_like "event data"
```
martin308: it would be great to include the http fields checks here as well! We would probably need to include some additional headers in a `before` block to satisfy the conditions though
```suggestion
      it_behaves_like "event data", http_fields: true
```

ajvondrak:
> I'm super curious what the code is actually trying to do? Why do we randomly grab a client[?]

The aws-sdk integration patches the common core object that all the actual AWS clients go through underneath. The test is essentially trying to smoke test that any arbitrary SDK client will still gain the powers bestowed by the beeline integration. But AWS has a lot of clients (191, as you can see below), so I didn't think it'd be a good idea to go through them exhaustively. Hence `Array#sample`.

> why do some of them require parameters that others don't?

The SDK clients are all for very different services. I guess the business of what parameters are/aren't required is up to Amazon. 🤷‍♂ 

Here's what I can find:
```ruby
clients = Aws::Partitions.service_ids.keys.map do |service|
  Aws.const_get(service).const_get(:Client)
end

clients.count
#=> 191

clients.select do |client|
  client.new
  false
rescue => e
  true
end
#=> [Aws::CloudSearchDomain::Client, Aws::IoTDataPlane::Client]

Aws::CloudSearchDomain::Client.new
#=> NameError: no member 'regional_endpoint' in struct

Aws::IoTDataPlane::Client.new
#=> ArgumentError: missing required option `:endpoint'
```

PR #75, opened 2020-03-07
Serializes ActionController::Parameters correctly

lizthegrey:
Don't prematurely call to_s on hashes, as we can actually serialize them in a way we can parse.

This enables us to construct nested JSON in libhoney, serializing things so they can be later unpacked on Honeycomb's side.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     honeycomb-beeline (1.3.0)
-      libhoney (~> 1.8)
+      libhoney (~> 1.14, >= 1.14.2)
 
 GEM
   remote: https://rubygems.org/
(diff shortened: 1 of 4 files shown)


lizthegrey:
> Can we add some tests for this, they should be pretty easy to add to [active_support_spec](https://github.com/honeycombio/beeline-ruby/blob/master/spec/honeycomb/integrations/active_support_spec.rb).
Sure thing.

> I'm wary of including the cleaner here as it was only added in `v1.14.1` so if someone is using an older version of libhoney this will break, also libhoney will be calling `clean_data` on the event data at some point so I'm not sure what the value of doing it here is. Maybe some tests will clarify this?
There isn't the ability in ruby to version beeline to depend upon a specific libhoney version? I'm totally not a ruby programmer...

As far as why: because clean_data is an easily reusable piece of code that extracts data types and serializes them to the appropriate JSON-encodable type. I started off trying to write something that would recognize hashes, arrays, etc. but then realized that it would be better to just reuse the same code. We could just pass the objects directly through without doing any kind of serialization, but then the objects might live until the next libhoney batch rather than being destroyable immediately after the call to the watcher...

PR #82, opened 2020-04-19
Add request.header.accept_language field

timcraft:
This would be useful for debugging locale related production issues.


paulosman:
@timcraft can you fix the failing tests? I think you just need to add the `header()` call to `spec/honeycomb/integrations/rails_spec.rb`

timcraft:
@paulosman I left it because the tests were passing locally and the CI failures appeared to be unrelated, but yes... I've rebased the commit and added that call to the rails spec 👍

Closing in favour of #94 because I can't edit this pull request after having deleted the fork.

PR #85, opened 2020-05-08
Properly pass options for Ruby 2.7

terracatta:
In Ruby 2.7 passing the last argument as an options hash if the receiver uses keyword arguments is no longer allowed and emits a deprecation warning.

This PR applies the recommended fix of adding ** to covert the options hash into the correct keyword arg format.

There may be more deprecation warnings to fix in other parts of the codebase but these were the ones that jumped out to me right away.

### DIFF
--- a/lib/honeycomb/configuration.rb
+++ b/lib/honeycomb/configuration.rb
@@ -33,7 +33,7 @@ module Honeycomb
 
       @client ||
         (debug && Libhoney::LogClient.new) ||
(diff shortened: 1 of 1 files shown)


martin308:
@terracatta thanks! Looks like pry-byebug dropped support for ruby2.3 in a minor version bump :rage1: I'll push a PR to fix that

martin308:
@terracatta if you rebase this from master now we should get the tests running properly

PR #88, opened 2020-05-12
Fix regex substitution for warden and `empty?` errors for Rack

irvingreid:
Three things I found in testing the Rails beeline with our application:

- when finding available Warden scopes for user sessions, the regex substitution that was meant to extract the capture group containing the scope name was instead replacing it with a string containing the (invalid Unicode) character "\0001"
- The Rack integration throws an error if the value passed does not implement `empty?()`; Ruby only implements that method on collection types.
- Fields with boolean values would be omitted from the trace if the boolean was false, rather than being included with the `false` value. If this is intentional, fine, though I prefer to include it; there's a meaningful difference between "I don't have a value for that field" and "I have the value `false` for that field".


lib/honeycomb/integrations/warden.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module Honeycomb
-  # Methods for extracing common warden/devise fields from a rack env hash
+  # Methods for extracting common warden/devise fields from a rack env hash
```
paulosman, line 4: Nice. Thank you :-)

lib/honeycomb/integrations/rack.rb
```diff
@@ -31,9 +31,9 @@ def call(env)
       hny = env["HTTP_X_HONEYCOMB_TRACE"]
       client.start_span(name: "http_request", serialized_trace: hny) do |span|
         add_field = lambda do |key, value|
-          next unless value && !value.empty?
-
-          span.add_field(key, value)
+          unless value.nil? || (value.respond_to?(:empty?) && value.empty?)
```
martin308, line 34: Nice find, it would be great to have a test for this!

spec/honeycomb/integrations/rack_spec.rb
```diff
@@ -51,7 +51,7 @@ def last_name
       end
 
       def created_at
-        Time.now
+        Time.new 2002, 3, 4, 5, 6, 7
```
martin308, line 54: what was this fixing?

lib/honeycomb/integrations/rack.rb
irvingreid: The existing test covers both the regex match (we don't get any "user.<whatever>" values without that fix) and the `.respond_to?(:empty)` check (we get a `NoMethodError: undefined method 'empty?' for 2002-03-04 05:06:07 -0500:Time` error). The missing test related to my change is for `value == false` (though the Warden integration could also use tests for the logic around choosing which context to use, too...)

spec/honeycomb/integrations/rack_spec.rb
irvingreid: No specific issue, just a general aversion to having variable dates in tests; I've fixed too many tests that flaked out due to time quirks. I can back it out if you want.

lib/honeycomb/integrations/rack.rb
martin308: Ah great! That makes total sense

irvingreid:
Do you still want two separate PRs or is this one OK?

martin308:
@irvingreid I think we could pull one or two small things out but they are much smaller than I initially thought so let's just use this one. Sorry for the confusion and thanks for bearing with us

PR #91, opened 2020-06-04
Adding X-Forwarded-For to instrumented fields

paulosman:
For any apps using the rack integration, the X-Forwarded-For header will now be automatically instrumented. This is useful for services that are behind proxies, where remote_addr will likely be the proxy IP.


mrchucho:
You might consider adding `HTTP_X_FORWARDED_PORT`, too, since you've got protocol.

paulosman:
> You might consider adding `HTTP_X_FORWARDED_PORT`, too, since you've got protocol.

Yeah, it's not used as often, but I suppose it won't be instrumented if it's not present. I'll add it.

PR #92, opened 2020-06-04
Support custom notifications based on a regular expression

mrchucho:
ActiveSupport::Notifications support subscribing to regular
expressions, the beeline should too. Some internal code was using
the event name as a lookup key, so regular expressions weren't being
evaluated to match event names. This PR adds fallback logic: if we
can't look up by event name, try _matching_. Worst case, things fail
as they do now.


mrchucho:
> Can we get a test that shows the interaction we expect when you add a string and a regex that match the same key?

What is the expected interaction? Is it the same as if you had the same string key twice?

For example, if I subscribe to the same string key twice I get two events (same as w/ a regex):

```ruby
        [event_name, event_name].each do |event|
          subscriber.subscribe(event) do |_name, span, payload|
            payload.each do |key, value|
              span.add_field(key, value)
            end
          end
        end

        ActiveSupport::Notifications.instrument event_name, "honeycomb" => 1 do
        end
# ...
      it "sends a single event" do
        expect(libhoney_client.events.size).to eq 1
      end
```
```
  1) Honeycomb::ActiveSupport custom notifications with regex and string for same key sends a single event
     Failure/Error: expect(libhoney_client.events.size).to eq 1
     
       expected: 1
            got: 2
```

martin308:
Yup that’s it, would just be nice to have that expected behavior documented in a test

PR #96, opened 2020-06-12
Fix missing response code bug

piecehealth:
I found honeycomb didn't trace some 500 responses because it's set `response.status_code` after `status, headers, body = app.call(env)`, so if the call raise error, honeycomb client can't set `response.status_code`.

Usually, by this situation, the server would return 500 response, it's very important for honeycomb to be aware of this since by my understanding honeycomb should capture tracings according to response code.

### DIFF
--- a/lib/honeycomb/integrations/rack.rb
+++ b/lib/honeycomb/integrations/rack.rb
@@ -45,16 +45,19 @@ module Honeycomb
         span.add_field("request.secure", req.ssl?)
         span.add_field("request.xhr", req.xhr?)
 
-        status, headers, body = app.call(env)
+        begin
+          status, headers, body = app.call(env)
 
-        add_package_information(env, &add_field)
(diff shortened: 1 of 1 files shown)


martin308:
The event should always be sent if there is an error it's just that we can't automatically capture the status code in this case.

I'm not sure I like the idea of providing potentially correct information (we don't know the status code is going to be a 500) as part of the automatic instrumentation

Is there some other way this can be achieved where we can know the actual status code?

martin308:
@piecehealth one suggestion that we have come up with is that you could use a `presend_hook` to add this kind of logic to your application. An example of using one is over here https://github.com/honeycombio/beeline-ruby/blob/master/examples/hooks/main.rb#L25. You could check for a missing `response.status_code` field and the presence of `error` to set the status code to 500

piecehealth:
Thanks @martin308 , it works perfect for me

PR #98, opened 2020-07-01
remove children after sending

### LINKED ISSUE #84: Span should remove itself from it's parents children array after being sent
As mentioned in #83 we are not removing sent spans from the parents children array which causes them to be retained for the life of the parent span. 

The go beeline does this [here](https://github.com/honeycombio/beeline-go/blob/master/trace/trace.go#L321)

martin308:
Resolves #84.

The spec changes look really bad, but the main change is to dump the existing specs into another `describe` block and add an additional spec to check that after the child is sent it won't try to send any sent children. Might be easiest to just look at the last two commits as the shuffling around is isolated to the first commit.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -29,6 +29,9 @@ Metrics/LineLength:
   Exclude:
     - spec/support/event_data_shared_examples.rb
 
+Metrics/ParameterLists:
+  Max: 6
+
 Style/FrozenStringLiteralComment:
   EnforcedStyle: always
   Exclude:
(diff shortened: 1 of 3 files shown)


rintaun:
we applied this fix and it does appear to fix our problem. the memory usage graph mostly speaks for itself, I think:

image

martin308:
@rintaun thanks for testing this out! I'll get someone on @honeycombio/integrations-team to review this and we can make up a release

PR #100, opened 2020-07-22
UUID to w3c-friendly ids

katiebayes:
Just looking at switching these uuids over to w3c-standard trace ids (16-byte for trace ids, 8-byte for span ids). This might be out there to have a dependency for generating ids like this, but looking at the ruby otel package, the id generation was so nice and cleanly exported I figured it was worth trying.

We can also definitely just generate the bytes ourselves using the random package. Let me know what you think!

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -47,7 +47,6 @@ GEM
       addressable (~> 2.0)
       http (>= 2.0, < 5.0)
     method_source (0.9.2)
-    opentelemetry-api (0.5.1)
     overcommit (0.46.0)
       childprocess (~> 0.6, >= 0.6.3)
       iniparse (~> 1.4)
@@ -115,7 +114,6 @@ DEPENDENCIES
   bump
   bundler
   honeycomb-beeline!
-  opentelemetry-api
   overcommit (~> 0.46.0)
   pry (< 0.13.0)
   pry-byebug (~> 3.7.0)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -47,6 +47,7 @@ GEM
       addressable (~> 2.0)
       http (>= 2.0, < 5.0)
     method_source (0.9.2)
+    opentelemetry-api (0.5.1)
     overcommit (0.46.0)
       childprocess (~> 0.6, >= 0.6.3)
       iniparse (~> 1.4)
@@ -114,6 +115,7 @@ DEPENDENCIES
   bump
   bundler
   honeycomb-beeline!
+  opentelemetry-api
   overcommit (~> 0.46.0)
   pry (< 0.13.0)
   pry-byebug (~> 3.7.0)
--- a/honeycomb-beeline.gemspec
+++ b/honeycomb-beeline.gemspec
@@ -52,4 +52,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "simplecov"
   spec.add_development_dependency "simplecov-console"
   spec.add_development_dependency "webmock"
+  spec.add_development_dependency "opentelemetry-api"
 end
--- a/honeycomb-beeline.gemspec
+++ b/honeycomb-beeline.gemspec
@@ -42,7 +42,6 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "appraisal"
   spec.add_development_dependency "bump"
   spec.add_development_dependency "bundler"
-  spec.add_development_dependency "opentelemetry-api"
   spec.add_development_dependency "overcommit", "~> 0.46.0"
   spec.add_development_dependency "pry", "< 0.13.0"
   spec.add_development_dependency "pry-byebug", "~> 3.7.0"
--- a/honeycomb-beeline.gemspec
+++ b/honeycomb-beeline.gemspec
@@ -42,6 +42,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "appraisal"
   spec.add_development_dependency "bump"
   spec.add_development_dependency "bundler"
+  spec.add_development_dependency "opentelemetry-api"
   spec.add_development_dependency "overcommit", "~> 0.46.0"
   spec.add_development_dependency "pry", "< 0.13.0"
   spec.add_development_dependency "pry-byebug", "~> 3.7.0"
@@ -52,5 +53,4 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "simplecov"
   spec.add_development_dependency "simplecov-console"
   spec.add_development_dependency "webmock"
-  spec.add_development_dependency "opentelemetry-api"
 end
--- a/lib/honeycomb/propagation.rb
+++ b/lib/honeycomb/propagation.rb
@@ -56,6 +56,7 @@ module Honeycomb
 
       value.downcase!
       Array(value).pack('H*')
+    end
 
     def parse_trace_id(string)
       INVALID_TRACE_ID = ("0" * 32).b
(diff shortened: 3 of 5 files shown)


honeycomb-beeline.gemspec
```diff
@@ -42,6 +42,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "appraisal"
   spec.add_development_dependency "bump"
   spec.add_development_dependency "bundler"
+  spec.add_development_dependency "opentelemetry-api"
```
martin308, line 45: this would have to use the `add_dependency` not `add_development_dependency` as it looks like it is used in the actual gem not just tests?

martin308: We would also need to specify a version that we are compatible with. Does it have a stable API yet?

katiebayes: good point!

katiebayes:
With the versioning question on the table (which we should def discuss but later in the week), I think it might make more sense to not use a gem here. Thanks for spotting that! We can always switch back over to a maintained gem later if we think it's a good move.

Gemfile.lock
```diff
@@ -114,6 +115,7 @@ DEPENDENCIES
   bump
   bundler
   honeycomb-beeline!
+  opentelemetry-api
```
martin308, line 118: we ned to revert or run bundler again to remove these lines

lib/honeycomb/span.rb
```diff
@@ -20,11 +20,20 @@ class Span
...
 
+    INVALID_SPAN_ID = ("\0" * 8).b
+
+    def generate_span_id
+      loop do
+        id = SecureRandom.hex(8)
+        return id unless id == INVALID_SPAN_ID
+      end
+    end
+
```
martin308, line 31: can we move all this below the `private` marker?

lib/honeycomb/trace.rb
```diff
@@ -17,11 +16,20 @@ class Trace
...
 
+    INVALID_TRACE_ID = ("\0" * 16).b
+
+    def generate_trace_id
+      loop do
+        id = SecureRandom.hex(16)
+        return id unless id == INVALID_TRACE_ID
+      end
+    end
+
```
martin308, line 27: can we move all these below this `private` marker?

lib/honeycomb/trace.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require "forwardable"
-require "securerandom"
```
martin308, line 4: this needs to be added back in

lib/honeycomb/trace.rb
```diff
@@ -34,5 +34,17 @@ def initialize(builder:, context:, serialized_trace: nil, **options)
     def add_field(key, value)
       @fields[key] = value
     end
+
+    private
+    
+    INVALID_TRACE_ID = ("\0" * 16).b
```
martin308, line 40: Thinking on this further, is this ever going to match something generated by `SecureRandom.hex`?

katiebayes:
Thanks @lizthegrey for talking me through the bytes vs hex question; it sounds like with otel using raw bytes internally, we should probably do the same and transform to hex on propagation only. I'll get working on that change

martin308:
@katiebayes that should resolve my question as well about wether the guard will be hit comparing `INVALID_TRACE_ID` with the call to `hex`

lib/honeycomb/trace.rb
katiebayes: Nope! But see comment below, I dug into the bytes discrepancy briefly and we should move this over to raw bytes anyway. Also it's kind of interesting that ruby otel does this comparison on id generation. More commonly seen, is that a parser called on an incoming header will check against this invalid trace or span id and return undefined if it's detected.

martin308: surely you'd have to do the comparison here or you could generate spans/traces with invalid ids that would only be discovered when trying to send them?

lib/honeycomb/propagation.rb
```diff
@@ -47,17 +47,35 @@ def parse_v1(payload)
 
       [trace_id, parent_span_id, trace_fields, dataset]
     end
+
+    private
+
+    def parse_span_id(string)
+      INVALID_SPAN_ID = ("0" * 16).b
```
katiebayes, line 54: maybe? This is a little lazy, but also checks directly against the invalid string in the w3c spec https://www.w3.org/TR/trace-context/#trace-id. We could also move the invalid ids out of `private` and unpack them, which is what they're doing in otel

lib/honeycomb/propagation.rb
```diff
@@ -47,17 +47,35 @@ def parse_v1(payload)
...
+      value.downcase!
+      Array(value).pack('H*')
   end
 
   # Serialize trace headers
   module PropagationSerializer
     def to_trace_header
       context = Base64.urlsafe_encode64(JSON.generate(trace.fields)).strip
       encoded_dataset = URI.encode_www_form_component(builder.dataset)
+      encoded_trace_id = trace.id.unpack('H*')
```
katiebayes, line 73: Just trying to follow the existing pattern here but we could also just do this inline when defining `data_to_propagate`

lib/honeycomb/trace.rb
katiebayes: yeah that's a good point. I should probably add this check to nodejs

lib/honeycomb/span.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
-require "securerandom"
 require "forwardable"
+require "securerandom"
```
katiebayes, line 4: Is this still needed to use `Random`?

katiebayes:
@martin308 switched this over to use raw bytes internally, just a couple lingering questions and want to make sure this is going in the right direction

lib/honeycomb/span.rb
martin308: doesn't look like it, 💥 it!

martin308: We can also blast it from `trace.rb` as well

lib/honeycomb/propagation.rb
martin308: I'm not sure what you mean by move them out of private? We should pull the constants out of the method at least, but leaving them private seems good

lib/honeycomb/propagation.rb
```diff
@@ -47,17 +47,35 @@ def parse_v1(payload)
...
+
+    def parse_span_id(string)
+      INVALID_SPAN_ID = ("0" * 16).b
+      raise "invalid span id" if string == INVALID_SPAN_ID
+
+      value.downcase!
+      Array(value).pack('H*')
+
+    def parse_trace_id(string)
+      INVALID_TRACE_ID = ("0" * 32).b
```
martin308, line 61: same here, we should pull this constant out of the method

lib/honeycomb/propagation.rb
```diff
@@ -47,17 +47,35 @@ def parse_v1(payload)
...
+      raise "invalid trace id" if string == INVALID_TRACE_ID
+
+      value.downcase!
```
martin308, line 64: where is value coming from here?

katiebayes: that was from me initially doing this work inline above, and then opting to pull out the logic and pushing it up too fast. Sorry! Fixed

lib/honeycomb/span.rb
katiebayes: removed this

PR #103, opened 2020-07-28
Add ruby 2.2 to the test matrix

martin308:
We claim to support ruby 2.2 in the `gemspec` yet we don't run any of our tests on this version of ruby.

Adds 2.2 to the matrix and excludes framework versions that don't run on 2.2


martin308:
cc: @martinjt

PR #104, opened 2020-08-04
refactor parsers and propagators, add w3c and aws support

katiebayes:
(as part of improvements to support trace interoperability)

- moves honeycomb trace parsing/propagation to a namespaced module, in line with changes in other language beelines
  - for backwards-compatibility, includes the moved module in the existing one
- add case insensitivity on parsed header fields to honeycomb parser

For use in a later PR that will enable a configuration for custom parser/propagation hooks:
- adds aws parser and propagator
- adds w3c parser and propagator
- related test suites for both formats


lib/honeycomb/propagation.rb
```diff
@@ -4,63 +4,16 @@
...
-            rescue JSON::ParserError
-              trace_fields = {}
-            end
-          end
-        end
-      end
-
-      [trace_id, parent_span_id, trace_fields, dataset]
-    end
+    include HoneycombPropagation::UnmarshalTraceContext
```
katiebayes, line 12: We're going to be deprecating the non-namespaced parser/propagators, but for backwards compatibility this existing module should wrap the honeycomb-specified one it's been moved to.

lib/honeycomb/propagation/aws.rb
```diff
@@ -0,0 +1,67 @@
...
+
+require "base64"
+require "json"
+require "uri"
+
+module Honeycomb
+  # Parsing and propagation for AWS trace headers
+  module AWSPropagation
+    # Parse trace headers
+    module UnmarshalTraceContext
```
katiebayes, line 11: This is what we're calling similar things in other beelines so I threw the `Unmarshal/Marshal` here for parity, but I don't think this is absolutely necessary, as it hasn't been a term in this beeline before. Totally up for discussion

MikeGoldsmith: beeline-java uses `Encode` & `Decode` as part of the propagation codec design. I agree it's not totally necessary for it match others, but if we have the opportunity to synchronise, all the better for consistency 👍

lib/honeycomb/propagation/aws.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      def get_fields(fields)
+        trace_fields = {}
+        fields.each do |entry|
+          key, value = entry.split("=", 2)
+          case key.downcase
+          when "root"
+            trace_id = value
+          when "parent"
+            parent_span_id = value
```
MikeGoldsmith, line 38: We need prefer self over parent here, and only use parent if parent ID hasn't already been set.

```ruby
            parent_span_id = value if parent_span_id.nil?
```

lib/honeycomb/propagation/aws.rb
```diff
@@ -0,0 +1,67 @@
...
+      end
+    end
+
+    # Serialize trace headers
+    module MarshalTraceContext
+      def to_trace_header
+        context = ""
+        unless trace.fields.keys.nil?
+          trace.fields.keys.each do |key|
+            context.concat(";#{[key]}=#{trace.fields[key]}")
```
MikeGoldsmith, line 55: Wow - ruby's string interpolation looks _weird_!

lib/honeycomb/propagation/honeycomb.rb
```diff
@@ -0,0 +1,69 @@
...
+        end
+
+        [nil, nil, nil, nil]
+      end
+
+      def parse_v1(payload)
+        trace_id, parent_span_id, trace_fields, dataset = nil
+        payload.split(",").each do |entry|
+          key, value = entry.split("=", 2)
+          case key
```
MikeGoldsmith, line 32: Should we `key.downcase` here too to ignore casing?

lib/honeycomb/propagation/aws.rb
```diff
@@ -0,0 +1,67 @@
...
+          split = serialized_trace.split(";")
+
+          trace_id, parent_span_id, trace_fields = get_fields(split)
+
+          parent_span_id = trace_id if parent_span_id.nil?
+
+          trace_fields = {} if trace_fields.nil?
+
+          if !trace_id.nil? && !parent_span_id.nil?
+            return [trace_id, parent_span_id, trace_fields]
```
MikeGoldsmith, line 23: I think we need an extra `nil` for dataset compared to honeycomb's implementation and the generic return below. Although, I'm not sure it's strictly necessary because ruby is _magic_.

```suggestion
            return [trace_id, parent_span_id, trace_fields, nil]
```

lib/honeycomb/propagation/aws.rb
katiebayes: My intention here (though far from perfect, might make more sense as an `if`) was that putting `self` _below_ parent in a case statement will mean that `self` will overwrite anything set by `parent`. So if there's a `parent` and no `self`, `parent` will get assigned, same deal vice versa, and with both, the final assignment wins.

lib/honeycomb/propagation/honeycomb.rb
katiebayes: oooh I like that

lib/honeycomb/propagation/aws.rb
katiebayes: good point, and I think it can't hurt. I'll add this to both w3c and aws. Even if it wouldn't necessarily break, I think handling all 4 potential fields will help readability

lib/honeycomb/propagation/aws.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+require "base64"
+require "json"
+require "uri"
```
martin308, line 5: Are any of these `require` things used in this file?

lib/honeycomb/propagation/w3c.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+require "base64"
+require "json"
+require "uri"
```
martin308, line 5: Are any of these require things used in this file?

lib/honeycomb/propagation/w3c.rb
```diff
@@ -0,0 +1,53 @@
...
+              return [trace_id, parent_span_id, nil, nil]
+            end
+          end
+        end
+        [nil, nil, nil, nil]
+      end
+
+      def parse_v1(payload)
+        @invalid_trace_id = "00000000000000000000000000000000"
+        @invalid_span_id = "0000000000000000"
```
martin308, line 29: These could probably be constants in this file?

katiebayes:
> I also noticed these files are not required anywhere by the beeline, is that to come later?

Yes, I know it's kind of a weird process, but it's just to break it up a bit; create the parsers and get those available, and then I'll push up a separate PR to expose configuration "hooks" and the config option.

lib/honeycomb/propagation/w3c.rb
katiebayes: uhhh nope. Good catch :)

lib/honeycomb/propagation/w3c.rb
katiebayes: yes, I think I went overboard with the `@` variables so I've removed those, but let me know if I'm interpreting this change incorrectly!

lib/honeycomb/propagation/aws.rb
katiebayes: Nope, thanks for catching. Removing!

dsoo:
I don't think this sequencing is absolutely required, but I thought that the idea was that we were going to add the custom parser and propagation hooks first, THEN add w3c and aws? This adds these marshal and unmarshal functions without any code that actually uses them, at the risk we might decide to refactor them once we actually implement the custom hooks that use them?

katiebayes:
> I don't think this sequencing is absolutely required, but I thought that the idea was that we were going to add the custom parser and propagation hooks first, THEN add w3c and aws? This adds these marshal and unmarshal functions without any code that actually uses them, at the risk we might decide to refactor them once we actually implement the custom hooks that use them?

I'm not married to this order of happenings, but this is how it went with golang and node. But now that you mention this I think it makes a bit more sense to enable custom hooks first... The marshal/unmarshal functions aren't complete hooks either, so it's more to break up the process a bit. We can hold off on merging this and get the configuration option in first?

MikeGoldsmith:
We added the AWS & W3C propagation codec's before the parser and propagation hooks in beeline-java too. I don't think it's too important, as long as all the work is done between releases.

katiebayes:
Amaze! Just updated just to confirm the build checks, and will merge this in when we're green!

PR #105, opened 2020-08-11
fix(client): reference current span in start_span

rintaun:
Updates the Client#start_span method to store a reference to the current span, rather than repeatedly referencing `context.current_span`.

The method `Span#send` is a public method, which means that it is possible for instrumentation to send the span from within a block passed to `Client#start_span`. Under those circumstances, one of two problematic scenarios can occur:

1) If the current span is _not_ the root span, then start_span will end up sending the parent span instead of the child span.
2) If the current span _is_ the root span, then start_span will raise `ArgumentError: no method name given`. This is due to the fact that `context.current_span` returns `nil` and `NilClass#send` expects a parameter, i.e. the method name to send.

By storing a reference to the current span, we can guarantee that `Client#start_span` will always be operating on the correct span. Given that `Span#send` is idempotent, calling `#send` after a block has already sent the span is safe.

### DIFF
--- a/spec/honeycomb/client_spec.rb
+++ b/spec/honeycomb/client_spec.rb
@@ -158,14 +158,24 @@ RSpec.describe Honeycomb::Client do
 
   describe "sending from within a span block" do
     it "does not also send the parent span" do
-      root_span = client.start_span(name: "root")
-      client.start_span(name: "child", &:send)
-      expect(root_span.__send__(:sent?)).to be false
+      client.start_span(name: "root")
+
+      # rubocop:disable Style/SymbolProc
+      client.start_span(name: "child") do |child_span|
+        child_span.send
+      end
+      # rubocop:enable Style/SymbolProc
+
+      expect(libhoney_client.events.size).to eq 1
     end
 
     it "does not raise an error when the span is the root" do
       expect do
-        client.start_span(name: "child", &:send)
+        # rubocop:disable Style/SymbolProc
+        client.start_span(name: "child") do |child_span|
+          child_span.send
+        end
+        # rubocop:enable Style/SymbolProc
       end.to_not raise_error
     end
   end
--- a/spec/honeycomb/client_spec.rb
+++ b/spec/honeycomb/client_spec.rb
@@ -155,4 +155,18 @@ RSpec.describe Honeycomb::Client do
 
     it_behaves_like "event data", package_fields: false
   end
+
+  describe "sending from within a span block" do
+    it "does not also send the parent span" do
+      root_span = client.start_span(name: "root")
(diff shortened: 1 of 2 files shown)


rintaun:
Looks like the build is failing due to rubocop:

```
lib/honeycomb/client.rb:47:5: C: Metrics/MethodLength: Method has too many lines. [26/25]
    def start_span(name:, serialized_trace: nil, **fields) ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I can get it to `23/25` by changing the final `if block_given?` to a guard clause instead of a nested if/else. not sure how you feel about that

rintaun:
added two test cases showing the two bug scenarios described in the initial commit

rintaun:
* Reformatted `Client#start_span` to reduce method length
* Reformatted test cases to use `&:send`... I'm not actually sure about this change. Personally, I find the actual block calling `child_span.send` to be much clearer in what is actually going on...

rintaun:
the `faraday_0` gemfile is failing, but I don't get that failure locally with that gemfile

rintaun:
Just to document the concrete scenario that led to encountering this bug, we were instrumenting forked Resque workers. Due to how it's set up, we were ensuring that we were sending the trace prior to the forked process exiting; however, in _some_ cases that occurs within another span (which we don't necessarily have control over), and sending in that case causes the parent span to be sent as well, as illustrated in the tests.

As far as we're aware, it didn't cause any actual problems (aside from span durations being cut slightly short, perhaps), but that might not always be the case.

spec/honeycomb/client_spec.rb
```diff
@@ -155,4 +155,18 @@
 
     it_behaves_like "event data", package_fields: false
   end
+
+  describe "sending from within a span block" do
+    it "does not also send the parent span" do
+      root_span = client.start_span(name: "root")
+      client.start_span(name: "child", &:send)
+      expect(root_span.__send__(:sent?)).to be false
```
martin308, line 163: Is there a specific reason to use `__send__` that I'm not seeing? If not lets check the events that libhoney is told to send rather than use the `__send__` here.

something like:

`expect(libhoney_client.events.size).to eq 1`

rintaun: Good call. I had that thought about the private method call last night, then forgot about it 😳

I just pushed a new commit with your suggestion. My other thought was `expect(client.current_span).to eq root_span`, but I think they are effectively checking the same thing.

I also reformatted the `&:send` back to block form and disabled that cop for those blocks.

PR #110, opened 2020-09-09
[draft-wip] add custom parse and propagation hook configs

katiebayes:
Enable custom hook configurations for trace header parsing and propagation.
- similar to the other beelines, the `propagation_context` hash will be used as the intermediary data structure for both parsing and propagation. i.e. propagation hooks expect access to a propagation context for the current span at the time of the outbound request, and parsing hooks are required to return a `propagation_context`

### DIFF
--- a/lib/honeycomb/span.rb
+++ b/lib/honeycomb/span.rb
@@ -41,6 +41,7 @@ module Honeycomb
                       is_root: parent_id.nil?,
                       sample_hook: nil,
                       presend_hook: nil,
+                      http_trace_propagation_hook: nil,
                       **_options)
       @parent = parent
       # parent_id should be removed in the next major version bump. It has been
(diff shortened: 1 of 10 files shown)


lib/honeycomb/span.rb
```diff
@@ -36,11 +36,12 @@ def initialize(trace:,
       parse_options(**options)
     end
 
-    def parse_options(parent: nil,
+    def parse_options(parent: nil, # rubocop:disable Metrics/ParameterLists
```
katiebayes, line 39: I opted to disable rubocop on this line rather than jump through the hoops to refactor or group the parameters, mainly because it looks like we'll be shaving one param back off here in the next major version. Lemme know what you think though, I know it's cheating a little :)

lib/honeycomb/trace.rb
```diff
@@ -17,9 +17,22 @@ class Trace
...
+      propagation_context: nil,
+      **options
+    )
+
+      # if a propagation context hash was successfully created, use that
+      # otherwise fall back to previous behavior of parsing a string
+      if propagation_context
+        trace_id, parent_span_id, trace_fields, dataset =
+          propagation_context
+      else
```
katiebayes, line 32: keeping this backwards compatible for now

PR #112, opened 2020-09-16
add custom config options for trace parsing and propagation

katiebayes:
(creating an org branch for easier review)

Currently going through this a little more to add some notes and other options I'm playing with.

**Acceptance criteria as I'm currently understanding (but please edit as-needed):**
- [x] users can pass custom function for parsing incoming headers
  - [x] have control over which header keys will be read
- [x] users can pass custom function for formatting "propagation" outbound headers
  - [x] users can access propagation_context for current Span instance in their custom config function
- [x] instrumentations without any custom hooks fall back to honeycomb parsing and propagation, no error
- [x] consistent propagation_context object exists as intermediary for both propagation directions
  - *met, currently the same array and ordering that was in use for parsing, open to another way of solving this but concerned about backwards compatibility
- **tests**
  - [x] update parse and serialization tests
  - [ ] test that configurations are being applied and parsed in the respective libraries (Rack, Faraday)
    - not sure how best to test the behavior of the library (i.e. Rack) _with_ the parsers, but seems like we should do this

### DIFF
--- a/lib/honeycomb/client.rb
+++ b/lib/honeycomb/client.rb
@@ -10,7 +10,7 @@ module Honeycomb
   class Client
     extend Forwardable
 
-    attr_reader :libhoney, :propagation_hooks
+    attr_reader :libhoney, :parser_hook
 
     def_delegators :@context, :current_span, :current_trace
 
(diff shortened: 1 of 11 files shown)


lib/honeycomb/client.rb
```diff
@@ -35,20 +35,27 @@ def initialize(configuration:)
       @additional_trace_options = {
         presend_hook: configuration.presend_hook,
         sample_hook: configuration.sample_hook,
+        propagation_hook: configuration.http_trace_propagation_hook,
+        parser_hook: configuration.http_trace_parser_hook,
       }
 
+      @parser_hook = @additional_trace_options[:parser_hook]
```
katiebayes, line 42: This is slightly awkward but `additional_trace_options` go through a few modules before landing in `Span`. Most of the trace-involved actions seem to assume the current span to be the active `self` in scope. 

For context, Rack is currently where the first `client.start_span` is called. If we're going to parse in `Rack` as we discussed, we need this hook available to Rack. I guess we could also just call it directly in `client` though? I'll play with that

martin308:
I think this is covered by #117 now? We can reopen if not

PR #113, opened 2020-09-17
Custom hooks with classes

katiebayes:
(still figuring some things out around effective tests, but I believe the code is in a good place for review and collab)
Kinda started over quickly with a new tactic after taking a step back, this is mostly working now but with classes!
I'm wrapping the existing parse/propagation modules in these classes, but this seems like it will have more straightforward usage. Config hooks can look like this:
```
config.http_trace_parser_hook do |env|
  parser = Honeycomb::W3CPropagation::Parser.new
  parser.unmarshal_trace_context(env)
end
```

and 

```
config.http_trace_propagation_hook do |propagation_context|
    propagator = Honeycomb::W3CPropagation::Propagator.new
    propagator.marshal_trace_context(propagation_context)
end

```

Other notes on this version:
- using the `Client` as a landing place for checking if configs exist and applying custom hooks. See the design doc for notes, but essentially, this seemed like a good fit as the config is already in scope and won't need to be passed from here elsewhere
- Iffyness: @additional_trace_options has the custom parser and propagator hooks. These do seem to need to be stored in a symbol (or named hash item) to be able to call them later when needed, but since @additional_trace_options gets passed into the Trace and Span and the config won't be needed there with this setup, maybe a new variable for these?
- I hadn't thought to do this before but I do love having a `current_span_context` fetcher in `Context`. The context instance will be available in `Trace` and `Span` so that lends some nice flexibility

### DIFF
--- a/lib/honeycomb/client.rb
+++ b/lib/honeycomb/client.rb
@@ -48,7 +48,6 @@ module Honeycomb
     end
 
     def propagation_context_from_req(env:)
-      puts "propagation_context_from_req"
       custom_hook = @additional_trace_options[:parser_hook]
       if custom_hook.nil?
         parser = Honeycomb::HoneycombPropagation::Parser.new
@@ -58,11 +57,15 @@ module Honeycomb
       end
     end
 
-    def start_span(name:, serialized_trace: nil, **fields)
+    def start_span(
+      name:, serialized_trace: nil, propagation_context: nil,
+      **fields
+    )
       if context.current_trace.nil?
         Trace.new(serialized_trace: serialized_trace,
                   builder: libhoney.builder,
                   context: context,
+                  propagation_context: propagation_context,
                   **@additional_trace_options)
       else
         context.current_span.create_child
--- a/lib/honeycomb/client.rb
+++ b/lib/honeycomb/client.rb
@@ -48,23 +48,33 @@ module Honeycomb
     end
 
     def propagation_context_from_req(env:)
-      custom_hook = @additional_trace_options[:parser_hook]
-      if custom_hook.nil?
-        parser = Honeycomb::HoneycombPropagation::Parser.new
-        parser.unmarshal_trace_context(env)
-      else
-        custom_hook.call(env)
+      # set default parser to honeycomb
+      parser = Honeycomb::HoneycombPropagation::Parser.new
+      parser_hook = lambda do |req|
+        parser.http_trace_parser_hook(req)
       end
+
+      # if there's a custom hook, overwrite parser_hook with it
+      unless @additional_trace_options[:parser_hook].nil?
+        parser_hook = lambda @additional_trace_options[:parser_hook]
+      end
+
+      parser_hook.call(env)
     end
 
     def header_from_propagation_context(propagation_context)
-      custom_hook = @additional_trace_options[:propagation_hook]
-      if custom_hook.nil?
-        propagator = Honeycomb::HoneycombPropagation::Propagator.new
-        propagator.marshal_trace_context(propagation_context)
-      else
-        custom_hook.call(propagation_context)
+      # set default propagator to honeycomb
+      parser = Honeycomb::HoneycombPropagation::Propagator.new
+      propagation_hook = lambda do |context|
+        parser.http_trace_propagation_hook(context)
       end
+
+      # if there's a custom hook, overwrite propagation_hook with it
+      unless @additional_trace_options[:parser_hook].nil?
+        propagation_hook = lambda @additional_trace_options[:propagation_hook]
+      end
+
+      propagation_hook.call(propagation_context)
     end
 
     def start_span(
--- a/lib/honeycomb/client.rb
+++ b/lib/honeycomb/client.rb
@@ -47,7 +47,7 @@ module Honeycomb
       end
     end
 
-    def propagation_context_from_req(env)
+    def propagation_context_from_req(env:)
       if @additional_trace_options[:parser_hook].nil?
         parser = Honeycomb::HoneycombPropagation::Parser.new
         parser.unmarshal_trace_context(env)
--- a/lib/honeycomb/client.rb
+++ b/lib/honeycomb/client.rb
@@ -48,11 +48,13 @@ module Honeycomb
     end
 
     def propagation_context_from_req(env:)
-      if @additional_trace_options[:parser_hook].nil?
+      puts "propagation_context_from_req"
+      custom_hook = @additional_trace_options[:parser_hook]
+      if custom_hook.nil?
         parser = Honeycomb::HoneycombPropagation::Parser.new
         parser.unmarshal_trace_context(env)
       else
-        @additional_trace_options[:parser_hook].call(env)
+        custom_hook.call(env)
       end
     end
 
--- a/lib/honeycomb/client.rb
+++ b/lib/honeycomb/client.rb
@@ -56,7 +56,9 @@ module Honeycomb
 
       # if there's a custom hook, overwrite parser_hook with it
       unless @additional_trace_options[:parser_hook].nil?
-        parser_hook = lambda @additional_trace_options[:parser_hook]
+        parser_hook = lambda do |req|
+          @additional_trace_options[:parser_hook].call(req)
+        end
       end
 
       parser_hook.call(env)
@@ -71,7 +73,9 @@ module Honeycomb
 
       # if there's a custom hook, overwrite propagation_hook with it
       unless @additional_trace_options[:propagation_hook].nil?
-        propagation_hook = lambda @additional_trace_options[:propagation_hook]
+        propagation_hook = lambda do |context|
+          @additional_trace_options[:propagation_hook].call(context)
+        end
       end
 
       propagation_hook.call(propagation_context)
--- a/lib/honeycomb/client.rb
+++ b/lib/honeycomb/client.rb
@@ -62,7 +62,7 @@ module Honeycomb
       parser_hook.call(env)
     end
 
-    def header_from_propagation_context(propagation_context)
+    def headers_from_propagation_context(propagation_context)
       # set default propagator to honeycomb
       propagator = Honeycomb::HoneycombPropagation::Propagator.new
       propagation_hook = lambda do |context|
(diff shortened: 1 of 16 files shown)


lib/honeycomb/propagation/honeycomb.rb
```diff
@@ -53,16 +53,47 @@ def parse_v1(payload)
...
-        data_to_propogate = [
+      def to_trace_header(propagation_context: nil)
+        if propagation_context.nil?
+          trace_id = trace.id
+          span_id = id
+          trace_fields = trace.fields
+          dataset = builder.dataset
+        else
+          trace_id, span_id, trace_fields, dataset = propagation_context
+        end
```
katiebayes, line 64: This might be overkill, but it seemed appropriate to keep backwards compatibility for a while here. If `to_trace_header` is called by the `Propagator` class, it will have a `propagation_context`. But this module is still included in the Span class, and if called directly will fall back to expecting a `self` of a `Span` instance

lib/honeycomb/propagation/w3c.rb
```diff
@@ -42,14 +42,36 @@ def parse_v1(payload)
...
-      def to_trace_header
+      TRACE_ID_REGEX = /^[A-Fa-f0-9]{32}$/.freeze
+      SPAN_ID_REGEX = /^[A-Fa-f0-9]{16}$/.freeze
+
+      def to_trace_header(propagation_context: nil)
+        trace_id, span_id = propagation_context
         # do not propagate malformed ids
-        if trace.id =~ /^[A-Fa-f0-9]{32}$/ && id =~ /^[A-Fa-f0-9]{16}$/
-          return "00-#{trace.id}-#{id}-01"
+        if trace_id =~ TRACE_ID_REGEX && span_id =~ SPAN_ID_REGEX
```
katiebayes, line 51: this may be breaking since this is already exported. I think I'll add a fallback as a precaution

katiebayes: added!

lib/honeycomb/client.rb
```diff
@@ -44,11 +47,35 @@ def initialize(configuration:)
       end
     end
 
-    def start_span(name:, serialized_trace: nil, **fields)
+    def propagation_context_from_req(env:)
+      custom_hook = @additional_trace_options[:parser_hook]
```
dsoo, line 51: This logic still seems fiddly - Why are we not just setting a default custom hook to a honeycomb implementation, which just calls these two lines below? I remember discussing this, but don't remember if we came up with a reason to not do it. Also, it really feels like `custom_hook` should be `parser_hook`.

lib/honeycomb/client.rb
```diff
@@ -44,11 +47,35 @@ def initialize(configuration:)
...
+      if custom_hook.nil?
+        parser = Honeycomb::HoneycombPropagation::Parser.new
+        parser.unmarshal_trace_context(env)
+      else
+        custom_hook.call(env)
+      end
+    end
+
+    def header_from_propagation_context(propagation_context)
+      custom_hook = @additional_trace_options[:propagation_hook]
```
dsoo, line 61: Basically the same comments as above.

lib/honeycomb/client.rb
```diff
@@ -44,11 +47,35 @@ def initialize(configuration:)
...
+      if custom_hook.nil?
+        propagator = Honeycomb::HoneycombPropagation::Propagator.new
+        propagator.marshal_trace_context(propagation_context)
+      else
+        custom_hook.call(propagation_context)
+      end
+    end
+
+    def start_span(
+      name:, serialized_trace: nil, propagation_context: nil,
```
dsoo, line 71: Feels a bit weird that we're adding an optional `propagation_context` argument to every `start_span` call, instead of having a specific `propagate_and_start_trace` call for when propagation is happening.

For example, what should happen if you pass in a propagation context and there's already a trace? It feels like there are two explicit use cases, and this seems to be mixing them.

lib/honeycomb/context.rb
```diff
@@ -13,6 +13,15 @@ def current_span
       spans.last
     end
 
+    def current_span_context(span: current_span)
+      trace_id = span.trace.id
+      span_id = span.id
+      trace_fields = span.trace.fields
+      dataset = span.dataset
+
+      [trace_id, span_id, trace_fields, dataset]
```
dsoo, line 22: Really not a fan of returning a four element array here as opposed to a hash or a class instance. What happens if we want to add/remove/change this? Also, this is the propagation context, isn't it? Why is this called `current_span_context` instead of `propagation_context`?

lib/honeycomb/client.rb
```diff
@@ -44,11 +47,35 @@ def initialize(configuration:)
...
+      custom_hook = @additional_trace_options[:parser_hook]
+      if custom_hook.nil?
+        parser = Honeycomb::HoneycombPropagation::Parser.new
+        parser.unmarshal_trace_context(env)
+      else
+        custom_hook.call(env)
+      end
+    end
+
+    def header_from_propagation_context(propagation_context)
```
dsoo, line 60: Should this be plural, i.e. `headers_from_propagation_context`?

lib/honeycomb/propagation/aws.rb
```diff
@@ -47,20 +47,46 @@ def get_fields(fields)
 
     # Serialize trace headers
     module MarshalTraceContext
-      def to_trace_header
+      def to_trace_header(propagation_context: nil)
+        if propagation_context.nil?
```
dsoo, line 51: This feels wrong. We should ALWAYS have a propagation context - this assembly of data that SHOULD be in a propagation context from wherever it's coming from (globals, ???) is awkward - this means that this AWS implementation needs to know about our internals. I thought the general idea was that these hooks ONLY needed to know about propagation contexts.

Propagation context should be a required argument, and the function that calls `to_trace_header` should be requiring it.

If `to_trace_header` is a legacy API, then we should leave it alone and deprecate it - having it fill in a propagation context and then call the new propagation-context hook.

dsoo: Ah, saw your comment further down. This just generally applies to all `to_trace_header` implementations. In general for this sort of refactoring, I am strongly for reimplementing legacy interfaces in terms of new, cleaner, interfaces, rather than trying to accommodate both use cases with a single code path.

lib/honeycomb/propagation/w3c.rb
```diff
@@ -42,14 +42,41 @@ def parse_v1(payload)
...
+          return "00-#{trace_id}-#{span_id}-01"
         end
 
         nil
       end
     end
+
```
dsoo, line 63: I don't fully understand - if I am going to set a custom hook that does w3c propagation, how does one do that? There are no methods called `http_parser_hook` and `http_propagation_hook` like I would expect. In general, I don't fully understand why we have classes here at all.

dsoo (commented):
Half of this is, WTF, Ruby, why are you so weird? And is what we're doing really what people writing Ruby expect things to look like?

The other half of this is mostly about code clarity and structure. In general, I would seek to make sure that the "modern" code path looks clean and usable, and have legacy methods and codepoints reimplement in terms of the new code paths and perhaps be a bit uglier and less efficent, rather than have single code paths that handle both via branching on optional parameters, and trying to reuse existing implementations.

lib/honeycomb/client.rb
katiebayes: You're totally right, and I got caught up with all the trickiness in passing functions around. With the modules as they were before, which you can't really set as a namespaced default value so I gave up on trying to do that. Now that I'm using a class this should be more straightforward, I'll give it a whirl.

lib/honeycomb/context.rb
katiebayes: I know :( @paulosman I'm curious your thoughts here. I opted to keep this consistent with existing patterns but I don't love this being an array rather than a hash with named params. Worth changing?

lib/honeycomb/client.rb
katiebayes: @dsoo Ok! So it looks like the proper way to store a method in a variable is to use `lambda`. This way we can store that default, overwrite it with the custom hook if it exists, and then just call that variable as a method. We obviously still need the "is there a config?" logic somewhere but I think this is more straightforward and in line with what you're describing? Let me know what you think!

lib/honeycomb/client.rb
katiebayes: Yeah just thinking about this one. This is another one of those places where an existing flow and I mostly mimicked it, but I'm not sure the old flow is necessarily a problem? (I'm really not sure about this though.) The `Client.start_span` is kind of a misleading-ish name (probably could be `propagate_and_start_trace`, because the actual Span is initialized in `Trace` (which is initialized here). `Span` is of course the most central instance here, and the trace_id, parent_id, etc need to be available to all these instances without spinning up a new instance.

On that specific example, there's a check for `context.current_trace.nil?` before initializing a Trace (that receives a propagation_context), otherwise the `create_child` method is called on the current instance of `Span`.

lib/honeycomb/propagation/aws.rb
katiebayes: @dsoo Took a stab at refactoring this, let me know what you think: https://github.com/honeycombio/beeline-ruby/pull/113/commits/52472b27d8e4b2311ec9c82992c0160480a9455a

lib/honeycomb/propagation/w3c.rb
katiebayes: The configuration looks like this:
```
config.http_trace_parser_hook do |env|
    parser = Honeycomb::W3CPropagation::Parser.new
    parser.unmarshal_trace_context(env)
  end
```

but I just renamed some things to address this that I think will help. In the refactor above, I renamed `parse` to `unmarshal_trace_context` and renamed the above to `http_trace_parser_hook`. I think that should be more clear? So now the config will look like this (for a w3c parse):
```
  config.http_trace_parser_hook do |env|
    parser = Honeycomb::W3CPropagation::Parser.new
    parser.http_trace_parser_hook(env)
  end
```

paulosman:
Closing this - moving to #117.

PR #115, opened 2020-09-22
Add configuration for excluding child spans when a parent span is sam…

DanielHeath:
While not able to support interprocess traces, sampling all child spans makes it easy to decide a toplevel trace is 'not interesting' and discard all data associated with it.

### DIFF
--- a/lib/honeycomb/client.rb
+++ b/lib/honeycomb/client.rb
@@ -35,10 +35,10 @@ module Honeycomb
       @additional_trace_options = {
         presend_hook: configuration.presend_hook,
         sample_hook: configuration.sample_hook,
+        sample_excludes_child_spans: configuration.sample_excludes_child_spans,
         parser_hook: configuration.http_trace_parser_hook,
         propagation_hook: configuration.http_trace_propagation_hook,
       }
-
       configuration.after_initialize(self)
 
       at_exit do
--- a/lib/honeycomb/configuration.rb
+++ b/lib/honeycomb/configuration.rb
@@ -9,6 +9,7 @@ module Honeycomb
     attr_accessor :write_key,
                   :dataset,
                   :api_host,
+                  :sample_excludes_child_spans,
                   :debug
 
     attr_writer :service_name, :client, :host_name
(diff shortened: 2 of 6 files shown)


DanielHeath:
Ahh - I had an old copy around that was still using `master`, so I didn't pick up changes. Will rebase.

MikeGoldsmith:
Hi @DanielHeath - sorry for the slow reply. This seems to have slipped through out review net.

We'll review and get back to you soonn.

martin308:
@DanielHeath just reading over this now. One question that comes to mind is: what scenarios would this impact? The general flow is that children are sent first before the parent so I'm struggling to see what impact this would have.

DanielHeath:
The issue: I want to capture either an entire request, or nothing. Without this change, I get a smattering; 10% of one request, 30% of another.

The idea was that child spans would always have `parent` set, so I could propagate the sampling check.

One twist is that the parent is checked for sampling before it finishes - and might not even have all its event data set when deciding whether to sample. This is OK for my use since the parent span gets filled out before anything else happens.

I have been running it for a bit; some of it works, but not ActiveSupport notifications. 

Reading the code, I can see that the activesupport stuff doesn't set `parent`, so the sampling does not propagate. I should be able to fix that though...

DanielHeath:
Wait, nevermind. For this to work I need to store child spans to send later; otherwise you can't perform any action that generates a span while deciding whether to sample or not.

DanielHeath:
For that, I'd need to split out the concept of "span finished" and "span sent" over in context.rb...

DanielHeath:
Testing another iteration of this in our staging env now...

MikeGoldsmith:
Hi @DanielHeath - I just wanted to check-in to see if you've had the opportunity run any further tests?

DanielHeath:
image

I've still got the "We've had to throttle new data for this dataset" warning; once that passes from the account I'll be able to confirm whether it's working as I expect.

DanielHeath:
I've fixed up the merge conflicts on this. 2 more days before the throttle expires.
image

MikeGoldsmith:
Hi @DanielHeath - thanks for this contribution. However, we don't feel this is something we want to add to beeline-ruby. There are ways to achieve a similar result using OpenTelemetry Ruby using a custom sampler.

We’re going to close this for now. Feel free to re-open, if it’s still relevant.

PR #118, opened 2020-10-06
Use nil for config dataset when empty

MikeGoldsmith:
An empty dataset can cause issues and should default to `nil`.

### DIFF
--- a/lib/honeycomb/configuration.rb
+++ b/lib/honeycomb/configuration.rb
@@ -6,9 +6,10 @@ module Honeycomb
   # Used to configure the Honeycomb client
   class Configuration
     attr_accessor :write_key,
-                  :dataset,
                   :api_host,
                   :debug
+
+    attr_reader :dataset
 
     attr_writer :service_name, :client, :host_name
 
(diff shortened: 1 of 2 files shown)


irvingpop (approved):
Thanks for the quick turnaround Mike!!

paulosman:
LGTM - but do we know why a number of tests are now failing?

irvingpop:
Ah yeah looks like all the aws3 tests are failing

martin308 (changes requested):
We probably need to add a lot more tests for dataset being nil in various places, I think most of the tests provide an actual dataset right now

martin308:
@paulosman looks like a breaking change in a minor version bump of the aws dynamo gem

MikeGoldsmith:
Thanks for fixing the AWS dynamo tests in #119.

I can add more testing around having a nil dataset. @martin308 can you give any advice on where to focus?

paulosman:
When building a URI with a `nil` dataset, I think libhoney will still interpolate it as an empty string.

MikeGoldsmith:
Closing in favour of https://github.com/honeycombio/libhoney-rb/pull/67

PR #123, opened 2020-11-07
Move Middleware to be before ActionDispatch::ShowExceptions

lirossarvet:
This all came out of a very long conversation in the Honeycomb Pollinators slack. A user with a fresh Rails app seemed to be getting `response.status_code` set to `nil` on any errors in production. In local development it seemed to work, though this confusion went back and forth. We went back and forth setting a lot of config options, and eventually realized that unhandled exceptions were making it to the `Honeycomb::Rack::Middleware`. These would hit at the [`app.call`](https://github.com/honeycombio/beeline-ruby/blob/main/lib/honeycomb/integrations/rack.rb#L50) section and invoke the [`Honeycomb::Client#start_span`'s rescue block](https://github.com/honeycombio/beeline-ruby/blob/main/lib/honeycomb/client.rb#L71), set `error` and `error_detail`, and bypass the status code setting. 

The reason for this eventually turns out that [`ActionDispatch::ShowExceptions`](https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/middleware/show_exceptions.rb#L35), so long as `config.action_dispatch.show_exceptions` is either set to `true` or `nil` (which it is by default in all environments except `test`), then it sets the rack app response to be the error code and body. Similarly _in `development` only_, [`ActionDispatch::DebugExceptions`](https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/middleware/debug_exceptions.rb#L60) (technically anywhere that sets `config.action_dispatch.consider_all_requests_local = true`, but by default that's only `development`), will also rescue errors and render a fancy page. If that config is set to `false`, however, as it is in production, the error is just re-raised.

By placing Honeycomb _below_ `ShowExceptions` in the stack, the only default chance we have of rescuing errors and rendering an error response is `DebugExceptions`. That middleware, however, only does so for "local" requests (e.g., in development), and does nothing in production except log an error and re-raise. Placing the Honeycomb Middleware _above_ `ShowExceptions`, though, allows the error response to be set and Honeycomb to accurately send along the status code.

This issue can be replicated in at least the `rails52` example in this repo:

1. Configure the `honeycomb.rb` write key
2. I like to edit the `BeesController#index` to be something like `Bee.cant_beelieve_this_method`
3. `bundle exec rails s`
4. Visit localhost:3000/bees a few times then look at honeycomb, grouping by `response.status_code`. They should all be set to 500. The page visited from the rails server should be the giant orange Rails error stack trace page.
5. Set `config.consider_all_requests_local` to `false` in `development.rb` then restart the server.
6. Visit localhost:3000/bees a few times then check Honeycomb, grouping by `response.status_code`. The will all be blank; the error page should be the "Something went wrong..." rails default page.

You can repeat the above with this patch and in _both_ cases you will get a `response.status_code` appropriately set.


lirossarvet:
Note that it was called out on https://github.com/honeycombio/beeline-ruby/pull/49 that perhaps this middleware should be after `Rails::Rack::Logger` which would've also avoided this. Given we already rely on `ShowExceptions` for the middleware placement, swapping `insert_after` for `insert_before` feels safe and backwards compatible for existing applications.

JustinC474:
👋 I am the aforementioned confused user 😉

Just wanted to voice a +1 for this PR. Currently, as @lirossarvet pointed out, on a standard Rails app you have 2 types of behavior for `http_request`s.

1. Everything is OK - the span has a `status_code`
2. There is an error - the span has no `status_code` but has a `error` and `error_detail`

(2) can span for 4xx and 5xx errors.

This change will align the behavior such that every `http_request` has a `status_code` on the span.

A side effect of this is that the `error` and `error_detail` will no longer be present on the 4xx and 5xx spans. If people want to capture those, they will need to add additional middleware.

However, this makes the behavior consistent for all `http_request`s, which I believe makes for a better product and a better user experience.

+1 to this change 👍

martin308 (changes requested):
This makes sense to me! I'd love to be able to have a test with this as it looks like the Railtie insertion of the middleware is a bit of a gap in the testing right now. Perhaps we could make the actual insertion of the middleware be a class method on `Honeycomb::Railtie` and then we could call that class method in `rails_spec.rb` instead of manually inserting the middleware. Does that sound ok?

lirossarvet:
Heh, I was afraid you'd say that. I'm going to try to test the Railtie specifically by defining it like so:

```ruby
module Honeycomb
  # Automatically capture rack requests and create a trace
  class Railtie < ::Rails::Railtie
    def self.insert_honeycomb_middleware(app)
      if Honeycomb.client
        app.config.middleware.insert_before(
          ActionDispatch::ShowExceptions,
          Honeycomb::Rails::Middleware,
          client: Honeycomb.client,
        )
      end
    end

    initializer("honeycomb.install_middleware",
                after: :load_config_initializers) do |app|
      self.class.insert_honeycomb_middleware(app)
    end
  end
end
```
then passing in a Rails application created _after_ the `require` runs.

I've never tested a Railtie directly, so please let me know if you think this scenario is misguided.

martin308:
Yeh I think that is along the right track, I was thinking more like this so we can pass the app and the client in which would allow the `rails_spec` test to more easily call it

```ruby
module Honeycomb
  # Automatically capture rack requests and create a trace
  class Railtie < ::Rails::Railtie
    initializer("honeycomb.install_middleware",
                after: :load_config_initializers) do |app|
      if Honeycomb.client
        self.class.insert_middleware(app: app, client: Honeycomb.client)
      end
    end

    def self.insert_middleware(app:, client:)
      app.config.middleware.insert_before(
        ActionDispatch::ShowExceptions,
        Honeycomb::Rails::Middleware,
        client: client,
      )
    end
  end
end
```

martin308:
Then replace this https://github.com/honeycombio/beeline-ruby/blob/main/spec/honeycomb/integrations/rails_spec.rb#L44-L48 with the call to the new class method

lirossarvet:
@martin308 ah yeah, that totally works. I pushed up a commit that addresses that; I initially hadn't thought of using the Railtie method to do that in the rails spec so thanks for the guidance!

martin308:
@lirossarvet no worries! Thanks for adding the changes. 

I have one more suggestion that I just thought of. Adding a test to the `rails_spec` file that has an error in the controller. This would involve adding an additional action to the controller in that file that just did a `raise` and then duplicating `describe "a standard request"` to call the error action. The shared examples in there should cover checking for the status code

Let me know if you want to take that one or I can push the change to your branch if not?

lirossarvet:
@martin308 Got it!

Let me know if you want to rebase this so you can merge in as a single commit or if you all have your own process for that. 

For fun, I also made sure that changing the ordering to match the Railtie in the `rails_spec.rb` file caused them to fail in `main` branch and they did and they pass now, so that feels good. Thanks for pushing for that test; I think that'll be good :)

martin308 (approved):
Perfect! We do a squash merge so this is all good

robbkidd:
In considering a release containing this change, do y'all think the disappearance of the `error` and `error_detail` fields on `http_request` spans is backwards-breaking?

PR #125, opened 2020-12-01
Support HTTP Accept-Encoding header

irvingreid:
This is helpful when trying to diagnose confusion over whether HTTP responses should be compressed.

To be honest it would be nice to get the `Content-Encoding` and `Vary` response headers too, but the Rack integration doesn't generally collect response headers.


irvingreid:
Hmm. This is an optional header, not sure how to test for that in your harness.

martin308:
@irvingreid if you rebase this then it should pick up the fixed tests in the `main` branch

irvingreid:
> @irvingreid if you rebase this then it should pick up the fixed tests in the `main` branch

@martin308 Thanks for following up. My git-fu may be failing me, but I don't see any new changes to rebase onto?

martin308:
@irvingreid 🤦 you're right! This was a fix to the AWS gem, I've hit re-run on the build

robbkidd (approved):
accept it

I don't immediately see a way to add an automated test for the absence of this header, but several of the other headers included are also optional, so I'm OK without. A quick trial run of the example Rails 5.2 app shows `request.header.accept_encoding` appearing in the `http_request` span when present in the request and not appearing when absent. 👍

PR #129, opened 2021-01-05
Add dependabot config

MikeGoldsmith:
Checks weekly and updates @honeycombio/integrations-team.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,7 +6,7 @@
 version: 2
 updates:
   - package-ecosystem: "bundler" # See documentation for possible values
-    directory: "/" # Location of package manifests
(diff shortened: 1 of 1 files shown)


martin308 (commented):
This is probably worth having to be consistent with the other repos. I'm not sure how much value we will get out of it though as we use the gemfiles in the gemfiles directory and we purposely don't check in lock files for those. 

Maybe we should set the directory property to gemfiles and see what happens?

MikeGoldsmith:
Yeah, sure - lets give that a try 👍

robbkidd:
Rebased branch to latest `main`.

robbkidd (approved):
Let's give it a whirl!

a cyberman dancing

PR #132, opened 2021-01-07
Add test case for 5xx response

vreynolds:
* adding back test coverage introduced in https://github.com/honeycombio/beeline-ruby/pull/123
* #123 was reverted so we can continue to release a non-major version off main


robbkidd (approved):
copies of tests

PR #134, opened 2021-01-07
Ensure status_code is set on the event, when there is an exception

vreynolds:
This brings back the change from #123 

* this will ensure status_code is always set on the event, even when it's a 4xx and 5xx response
* without additional work this will remove error and error_detail from event in those cases, making this a breaking change


MikeGoldsmith:
We’re going to close this for now. Feel free to re-open, if it’s still relevant.

PR #141, opened 2021-01-25
expand on the Rails 5.2 example

robbkidd:
### add a background job so the app pokes itself routinely
+ adds Sidekiq as the background job engine
+ adds sidekiq-cron to be able to schedule background jobs thereby causing the app to generate slightly interesting events to ship out to Honeycomb without human interaction with the app

+ adds a SchedulePokes job to put poking jobs on the queue
+ adds a PokeJob to perform an HTTP client request to the app
+ adds Faraday to use as the client for HTTP requests which will take advantage of the Faraday instrumentation to report data about the request

+ adds Honeykiq, a community-maintained library for instrumenting Sidekiq jobs
+ add a scheduled job to run Honeykiq's periodic reporter to collect Sidekiq stats

### make running the whole thing easier
+ add a Dockerfile to build a local image of the Rails app
+ arrange the Gemfile to retrieve the Beeline from RubyGems when building and running in Docker because the container will not have access to the filesystem two directories above here where the Beeline library lives in this repository

+ add a docker-compose config for running the multiple services:
    - the Rails web process
    - the Sidekiq job runner(s)
    - a Redis instance (dependency of Sidekiq)
    - a Squid proxy (to demonstrate proxy config via env vars and the reverse proxy back to network-isolated web app)

+ update the Honeycomb Rails initializer to demonstrate getting Honeycomb config values from environment variables


examples/rails52/.proxyconfig/htpasswd
```diff
@@ -0,0 +1 @@
+pxuser:$apr1$wr8E9m1.$Jyz5NrKJf7I3x19a2NYgH0
```
vreynolds, line 1: Is this a fake password?

robbkidd: Yep! This is for the `pxuser` used for authenticated forward proxying that is [set in the docker-compose.yml](https://github.com/honeycombio/beeline-ruby/pull/141/files#diff-191eb201a4025ad4612d188098339ec4c44e8ca770a4fe8aae539ef731c5da06R14). It is only used there in this wee standalone environment and not a _secret_ secret.

partounian:
Thanks for this! Definitely going to use recommendations to setup instrument sidekiq

vreynolds:
After `docker-compose up`, things _seem_ to be done but when I try to visit http://localhost:3000/ I keep getting 
```
ActiveRecord::PendingMigrationError (
 Migrations are pending. To resolve this issue, run:
         bin/rails db:migrate RAILS_ENV=development
```

paulosman (approved):
Tried it out and it worked :)

vreynolds:
My computer is allergic to rails 😢

robbkidd:
@vera Do you get past that error if you run `docker-compose up --build` instead to trigger a rebuild of the Rails app image? Could be there's a krufty previous image on your laptop. If that works, maybe I'll include the `--build` switch in the README steps because I've been including it just about every time I spin things up to make sure I'm using any of the latest edits.

vreynolds:
@robbkidd `docker-compose up --build` didn't fix it, but I can run `RAILS_ENV=development rake db:migrate` inside the rails container to get it to work. Are migrations supposed to run as part of container start?

robbkidd:
Yes! Migrations need to run, but it took be a great long time to realize why I hadn't needed to in a long while and why you did.

I've run the app locally on my laptop at least once
... which produced a `development.sqlite3` file locally with the migrations applied
... which has been included in every build of my local Rails app images ever since.

🤦 

I've just added the creation and migration of the database to the image build. Should be good enough for running this example. (No running multiple web containers then yet, though.)

vreynolds (approved):
🏅

PR #150, opened 2021-06-07
Switch from Thread to RequestStore

nvolker:
### Summary
Goodday my fellow tracing enthusiastic!

I noticed that the beeline is making use of `Thread.current` as its' context storage solution for the life of the request. During my time in the Ruby on Rails weeds, I've been exposed to some oddities and unreliabilities whilst using `Thread.current`. So, I have returned to offer a discussion in the form of a pull request. Unfortunately, I do not have any examples as I last debugged the beeline several months ago.

For more context on "The Problem" I'd like to link the **README.md** for the `RequestStore`:
https://github.com/steveklabnik/request_store#the-problem

With some added context here:
https://github.com/rails/rails/issues/8517#issuecomment-11438876

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,6 +3,7 @@ PATH
   specs:
     honeycomb-beeline (2.4.1)
       libhoney (~> 1.14, >= 1.14.2)
+      request_store
 
 GEM
   remote: https://rubygems.org/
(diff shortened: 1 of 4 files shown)


nvolker:
Closing this pull request given the gem's usage outside of the Rails context. The thread containing the discussion can be found on the Honeycomb Pollinators #sdk-ruby channel ([link here](https://honeycombpollinators.slack.com/archives/CJR134U2F/p1623079304039000)).

PR #151, opened 2021-06-08
Add ability to configure custom fields during Honeycomb configuration

nvolker:
### Summary
The ability to add custom fields during the configuration of a Honeycomb client can be extremely useful. Added flexibility can allow users to produce richer telemetry and enhance query capabilities. While this can be done during span creation, there are cases where a user may want to set additional fields during client creation which would, in turn, attach itself to **all** events.

An example, _for us_, is that we have a `sandbox` dataset where we'd like to further slice our data based on the "team" (given we have multiple sandboxes for teams to use). While this could also be achieved by configuring multiple datasets, there are architectural decisions that make it useful to have these data points present in the same dataset.

### Example Usage
```ruby
  config.dataset = "sandbox"
  config.service_name = "some-application-name"
  config.custom_fields = { team_name: "flow" }
  # ...
end
```

### If accepted, what's left?
- [ ] Specs
- [ ] Documentation Snippet

### DIFF
--- a/lib/honeycomb/client.rb
+++ b/lib/honeycomb/client.rb
@@ -30,6 +30,11 @@ module Honeycomb
 
       # maybe make `service_name` a required parameter
       @libhoney.add_field "service_name", configuration.service_name
+
+      configuration.custom_fields.each do |key, value|
+        @libhoney.add_field key, value
+      end
+
       @context = Context.new
 
       @additional_trace_options = {
--- a/lib/honeycomb/configuration.rb
+++ b/lib/honeycomb/configuration.rb
@@ -19,6 +19,7 @@ module Honeycomb
       @service_name = ENV["HONEYCOMB_SERVICE"]
       @debug = ENV.key?("HONEYCOMB_DEBUG")
       @client = nil
+      @custom_fields = {}
     end
 
     def service_name


ajvondrak:
You can drop a `Honeycomb.libhoney.add_field("team_name", "flow")` after the `Honeycomb.configure` block to accomplish the same thing. This will affect the global `Libhoney::Builder` that gets passed around the Beeline.

Example:

```ruby
require "bundler/inline"

gemfile do
  gem "honeycomb-beeline", source: "https://rubygems.org"
end

  config.debug = true
end

Honeycomb.libhoney.add_field("team_name", "flow")

Honeycomb.start_span(name: "parent") do
  Honeycomb.start_span(name: "child") do
    :ok
  end
end
```
```console
$ ruby example.rb 2>&1 | jq
{
  "meta.beeline_version": "2.4.0",
  "meta.local_hostname": "MBP",
  "meta.instrumentations_count": 10,
  "meta.instrumentations": "[\"active_support\", \"aws\", \"faraday\", \"rack\", \"rails\", \"railtie\", \"rake\", \"redis\", \"sequel\", \"sinatra\"]",
  "service_name": null,
  "team_name": "flow",
  "name": "child",
  "duration_ms": 0.008999952115118504,
  "trace.trace_id": "85b8408b3a0b6f48f6da4c5ebba092f3",
  "trace.span_id": "8aa06bbc7dbc6665",
  "meta.span_type": "leaf",
  "trace.parent_id": "a957913ca84e3f3c"
}
{
  "meta.beeline_version": "2.4.0",
  "meta.local_hostname": "MBP",
  "meta.instrumentations_count": 10,
  "meta.instrumentations": "[\"active_support\", \"aws\", \"faraday\", \"rack\", \"rails\", \"railtie\", \"rake\", \"redis\", \"sequel\", \"sinatra\"]",
  "service_name": null,
  "team_name": "flow",
  "name": "parent",
  "duration_ms": 0.17399998614564538,
  "trace.trace_id": "85b8408b3a0b6f48f6da4c5ebba092f3",
  "trace.span_id": "a957913ca84e3f3c",
  "meta.span_type": "root"
}
```

nvolker:
That's a great alternative, thanks @ajvondrak. I was originally looking at the source code for configured client:
```ruby
DEVELOPMENT [1] pry(main)> show-source Honeycomb::Client#add_field

From: /[redacted]/gems/honeycomb-beeline-2.4.1/lib/honeycomb/client.rb:80:
Owner: Honeycomb::Client
Visibility: public
Signature: add_field(key, value)
Number of lines: 5

def add_field(key, value)
  return if context.current_span.nil?

  context.current_span.add_field("app.#{key}", value)
end
```

The check for a `current_span` had me seeking an alternative solution.

Out of curiosity, what's your preference toward calling the `Honeycomb.libhoney` client directly instead of configuring something like the proposed `custom_fields` in the `Honeycomb.client` directly?

nvolker:
@ajvondrak the suggested approach has me running into the following issue when using Honeycomb locally (the LogClient):
```ruby
➜  some-app git:(some-branch) ✗ rails c
  .../honeycomb_beeline.rb:42:in `configure_client!': undefined method `add_event' for #<Libhoney::LogClient:0x00007fa0ffd55d20> (NoMethodError)
```

My configuration currently looks like this:
```ruby
  config.client = Libhoney::LogClient.new if Rails.env.development?
  ...
end

Honeycomb.libhoney.add_event "foo", "bar"
```

martin308:
@nvolker in that scenario I would probably create a suitable client for your environment before configuring the beeline and pass the constructed client into the configuration, maybe something like this:

```
case Rails.env
when "development"
 client = Libhoney::LogClient.new
else
  client = Libhoney::Client.new(writekey: 'YOUR_API_KEY',
                                dataset: 'honeycomb-ruby-example')
  client.add_field("wow", "wee")
end

  config.client = client
  ...
end
```

ajvondrak:
@nvolker The method is `#add_field`, not `#add_event`. 🙂 https://github.com/honeycombio/libhoney-rb/blob/a349c0e92a096ce1bbe69b1018e7fcd647e96aae/lib/libhoney/builder.rb#L47-L57

nvolker:
@martin308, that's an incredibly helpful comment. Thanks. I'm going to close this pull request for the time being 😺 Thanks you both!

PR #152, opened 2021-06-15
Allow setting different notification handling logic for specific events

lirossarvet:
NOTE: I still intend to write some tests for this! I just wanted to solicit feedback on the overall approach and if there's a desire for it first.

@ajvondrak and I were chatting on Pollinators about my desire to be able to customize how Honeycomb adds fields from the ActiveSupport::Notification payload. For example, `perform.active_job` passes a job object so when it gets serialized for tracing becomes `#<MyCoolJob:0x0000560f94402f28>` which is only so helpful. I would, personally, get more benefit if I could just add the class or even break it into distinct pieces, like pulling `enqueued_at` out or the job ID, or ignoring it outright :D 
If the desire is just to remove a field, this *can* be done with the `presend_hook` handler. However, that has a couple of small problems. 

1) it can only remove the data
2) `presend_hook` can quickly grow unwieldy. While other options exist that can help break it into more digestible chunks, such as a Hash (though I hazard that many people go for `case` blocks or `if`/`elsif` blocks:

```ruby
SCRUBBER_CLASSES = {
  "sql.active_record" => ActiveRecordFieldScrubber
}
SCRUBBER_CLASSES.default = DefaultFieldScrubber

config.presend_hook do |fields|
  SCRUBBER_CLASSES[fields["name"]].scrub(fields)
end
```

This is all well and good for scrubbing or modifying _based on what Honeycomb serializes_, but less so if you want to change how or what the raw data is serialized into.

Admittedly, again, I could implement the same or something similar as what is done in the ActiveSupport integration provided. However, it seems like it would be nice to let people piggyback off of the complex part you've all solved for us already and just do something else. I could even see a similar world to this PR where the custom handler function returns a Hash of fields that are then handled by the honeycomb integration. 

This isn't a make-or-break feature request by any means. It's just something I've encountered and thought "huh, wish I could change this, just a little." I'm happy to be told this is overthinking or not really a problem, or complexity for complexity's sake, which means I can happily run my fork or something if I desperately want to, touching no files but this one so I can always stay up to date.


lirossarvet:
Sidenote but related: it might be nice to include something like this in the docs for the ruby beeline to help people get a testable presend hook. Doesn't have to be what I had! Could be more barebones.

```ruby
class HoneycombPresendHookHandler
  def self.call(fields)
    SCRUBBER_CLASSES[fields["name"]].scrub(fields)
    
    # some other cleanup things, maybe other classes that deal with certain
    # data that gets added conditionally or something
  end
end
```

lirossarvet:
Shoot, also meant to post my little test script I threw together

```ruby
require "bundler/inline"

ENV["HONEYCOMB_INTEGRATIONS"] = "active_support"

gemfile do
  source 'https://rubygems.org'

  gem "activesupport", "~> 5.2"
  gem "honeycomb-beeline", git: "https://github.com/lirossarvet/beeline-ruby", branch: "adjust-activesupport-handlers"
end

  config.write_key = ENV["HONEYCOMB_WRITEKEY"]
  config.dataset = "rails-test"

  config.on_notification_event do |name, span, payload|
    span.add_field("override_default_handling.go", payload[:value])
  end

  config.notification_events = ["mine.test_event", "mine.whatever"]
  config.register_notification_handler("mine.other_event") do |name, span, payload|
    span.add_field("#{name}.nice", payload[:value].to_i(16))
  end
end

def test_event
  ActiveSupport::Notifications.instrument "mine.test_event", value: rand(100).to_s(16) do
    sleep 0.1
  end
end

def other_event
  ActiveSupport::Notifications.instrument "mine.other_event", value: rand(100).to_s(16) do
    sleep 0.1
  end
end

def default_handling
  ActiveSupport::Notifications.instrument "mine.whatever", value: rand(100).to_s(16) do
    sleep 0.1
  end
end

def unsubscribed
  ActiveSupport::Notifications.instrument "mine.never_subbed", value: rand(100).to_s(16) do
    sleep 0.1
  end
end

test_event
other_event
default_handling
unsubscribed

sleep 2
```

ajvondrak (commented):
Good work on the quick PR! 👀 

Looking over the implementation, I find it unintuitive that these new handlers are maintained separately. 🤔

Currently:
1. `config.notification_events` = subscribe to all of these events
2. `config.on_notification_event` = handle events (given above) with this block
3. If `config.on_notification_event` is missing, the default is to add each field from the `payload` to the `span` prefixed by `name`

The way this PR structures it, `config.register_notification_handler` subverts both configurations: it subscribes to separate events and uses a separate handler.

In my mind, this `register_*` hook is actually more similar to the `on_*` hook than it is different. It's like `on_notification_event` is "global", and `register_notification_handler` is parameterized on the event name (which I think also renders the block's `name` argument redundant). Makes me think they should actually be the same API.

So I was imagining it more like this:

```diff
diff --git a/lib/honeycomb/integrations/active_support.rb b/lib/honeycomb/integrations/active_support.rb
index fb362f6..91b5eff 100644
--- a/lib/honeycomb/integrations/active_support.rb
+++ b/lib/honeycomb/integrations/active_support.rb
@@ -21,17 +21,20 @@ def after_initialize(client)
         end
       end
 
-      def on_notification_event(&hook)
-        if block_given?
-          @on_notification_event = hook
+      def on_notification_event(name = nil, &hook)
+        @handlers ||= {}
+        if name
+          @handlers[name] = hook
         else
-          @on_notification_event
+          @default_handler = hook
         end
       end
 
       def handle_notification_event(name, span, payload)
-        if on_notification_event
-          on_notification_event.call(name, span, payload)
+        if @handlers && @handlers[name]
+          @handlers[name].call(span, payload)
+        elsif @default_handler
+          @default_handler.call(name, span, payload)
         else
           payload.each do |key, value|
             # Make ActionController::Parameters parseable by libhoney.
```

Then:

```ruby
  # This is still the exact set of events the handler subscribes to.
  config.notification_events = ["a", "b"]

  # This (optionally) overrides the default payload.each { span.add } handler.
  config.on_notification_event do |name, span, payload|
    span.add_field("#{name}.payload", payload)
  end

  # This takes precedence over the default handler (overridden or not)
  # specifically for the "a" event.
  config.on_notification_event("a") do |span, payload|
    span.add_field("a.payload", payload[:value])
  end
end
```
What do you think?

martin308 (commented):
I think I pretty much agree with @ajvondrak comments 👍

lirossarvet:
@ajvondrak @martin308 thanks for the feedback! I'll try to take a look later today or tomorrow.

robbkidd:
I'm intrigued. My interest will be in how the updated API feels and in test coverage for the customized handling.

robbkidd (changes requested):
I'm intrigued. My interest will be in how the updated API feels and in test coverage for the customized handling.

- Experiment with the shape of the API per @ajvondrak's recommendations
- Tests!
- Stretch goal: some draft examples of using the new customized handlers (at least in comments on the PR)

lirossarvet:
> thanks for the feedback! I'll try to take a look later today or tomorrow.

Okay so that obviously didn't happen. This is now my weekend project 😅 Sorry about the back and forth and delays on that.

robbkidd:
@lirossarvet Noooo worries! Thank you for your interest and engagement. It's OK to spend the time to make it feel good _and_ keep up with the rest of Life.

robbkidd:
@lirossarvet Also, you might be interested in rebasing against main. It's got some improvements to CI (clearer job names, hopefully better test result reporting, and judicious caching of dependencies) that could be helpful.

lirossarvet:
@robbkidd Okay! So I settled around most of @ajvondrak's API. The main difference is making the block `on_notification_event` accepts always use 3 arguments. Even though someone could in theory know the name of the instrumented event name, I think it's safer to keep those consistent. Additionally, someone might reuse a class and not really know what it is:

```ruby
class CoolStuff
  def self.sanitize(name, span, payload)
    payload.each do |k, v|
      span.add_field("#{name}.#{k}", v.to_s.gsub(/a-zA-Z/, '')
    end
  end
end

  config.on_notification_event('sql.active_record', CoolStuff.method(:call))
end
```

This is a terribly example, but it illustrates a possible use-case. Additionally, I made it so that you don't *have* to put specific events into the `notification_events` array.

I also did a little refactoring in the active support integration around only creating the default handler block once and using `Hash#fetch`; definitely let me know if that as out of bounds, though.

robbkidd:
> ... only creating the default handler block once and using `Hash#fetch`

😍   I love me some `Hash#fetch`. This is in the hopper for review soon.

ajvondrak:
> The main difference is making the block `on_notification_event` accepts always use 3 arguments. Even though someone could in theory know the name of the instrumented event name, I think it's safer to keep those consistent. Additionally, someone might reuse a class and not really know what it is:

Sensible enough. 👍

spec/honeycomb/integrations/active_support_spec.rb
```diff
@@ -133,7 +133,91 @@
...
+              hash_including("name" => "honeycomb.test_event", "the_hive" => "queen"),
+            ],
+          )
+        end
+        # rubocop:enable Metrics/LineLength
+      end
+    end
+
     describe "pass ActionController::Parameters as hash" do
+      require "action_controller"
```
ajvondrak, line 219: 🤨 Seems suspicious that this needs to be `require`d manually. Would it not have failed in CI?

spec/honeycomb/integrations/active_support_spec.rb
```diff
@@ -133,7 +133,91 @@
       end
     end
 
+    describe "custom notifications with event specific handler hook" do
+      let(:libhoney_client) { Libhoney::TestClient.new }
+      let(:configuration) do
+        Honeycomb::Configuration.new.tap do |config|
+          config.client = libhoney_client
+          config.notification_events = ["honeycomb.nonspecific_event"]
+          # rubocop:disable Metrics/LineLength
```
ajvondrak, line 142: If I'm picking nits, it seems like there's an awful lot of `rubocop:disable`s around this PR, especially since there are only 2 other disables in the rest of the code base:
```console
$ rg 'rubocop:disable'
spec/honeycomb/client_spec.rb
169:      # rubocop:disable Style/SymbolProc
180:        # rubocop:disable Style/SymbolProc
```
Seems like a signal that the code should be changed. But then, I'm a stickler for following rubocop - possibly more than others consider healthy. Also, I realize the lines are just gonna kind of run long with all the parameters n shit, so. 🤷

lirossarvet:
I'll rebase this again to get it up to date. If it all looks good and you're fine with the API, I'll write up the examples you asked for :) I just didn't want to do it then have to change the API again and make the examples again 👯

ajvondrak (commented):
Nooice

Looks solid. Dunno how others feel about `|`-ing together the `notification_events` and the custom handlers, but it's defensible.

spec/honeycomb/integrations/active_support_spec.rb
lirossarvet: Yeah, I am also usually a stickler for following rubocop more strictly than some. I also take it a little easier on things like LineLength or ClassLength in spec files, though. I felt like these would've been _less_ readable if split onto multiple lines somehow, especially with the wanting to keep the block params having useful names. However, happy to try to find a way to appease Rubocop without telling it to go away.

spec/honeycomb/integrations/active_support_spec.rb
lirossarvet: Heh. I don't think it'll fail CI because CI will run `BUNDLE_GEMFILE=rails_52.gemfile bundle exec rspec` and I was just running this _specific_ file which didn't include action controller. I am happy to remove it since it was mainly a "local testing just this file" kind of thing.

lirossarvet:
> Dunno how others feel about |-ing together the notification_events and the custom handlers, but it's defensible.

Totally hear you! Let me at least lay out my reasons. It seemed...weird to require someone to put it in a list of notification_events Honeycomb is tracking if they've already explicitly done `on_notification_event("event.name")`. Clearly they wanted to track it. I can change how we combine them if you're uncomfortable with the Array union. It is, in effect, the same as if I'd done:

```ruby
[ *notification_events, *active_support_handlers.keys].uniq
```

ajvondrak:
> I can change how we combine them if you're uncomfortable with the Array union

Oh, I'm fine with Array union. I was just more commenting on how `config.notification_events` isn't a "source of truth" if it gets amended elsewhere, that's all. You'd look at one config option and go "okay, I'm subscribing to events A, B, C" but another config option might be like "but also D, E, F". As you say, though, `config.on_notification_event(X)` is explicit in itself, so maybe parroting `X` in `config.notification_events` isn't really that worthwhile.

spec/honeycomb/integrations/active_support_spec.rb
ajvondrak: > I also take it a little easier on things like LineLength or ClassLength in spec files

For sure.

> I felt like these would've been _less_ readable if split onto multiple lines somehow, especially with the wanting to keep the block params having useful names

Yeah, I think that's probably right. And then `ActiveSupport::Notifications` doesn't help much with its already-long name. 😂

robbkidd: > ActiveSupport::Notifications doesn't help much with its already-long name.

Holy cats, it does not.

I nerfed the line length cop in a spec recently for readability, too, so I feel like we've got a precedent, at least for specs. I think I might prefer slightly that these be one-liner disables at the end of a long line (making it longer 🤦🏻 ) instead of multi-line, but ... not a blocker.

I'll allow it!

spec/honeycomb/integrations/active_support_spec.rb
robbkidd: Yea. Someday I might revisit how the multi-dimensional venn diagram of the test suite is wired, but today is not that day.

spec/honeycomb/integrations/active_support_spec.rb
lirossarvet: I honestly went back and forth on just wrapping the whole file in it or updating the rubocop.yml 😬 instead of having them scattered everywhere.

lirossarvet:
@ajvondrak @robbkidd You all seem Fine with the overall approach so far, so I'll put together some actual examples that look more realistic to what the imagined use case was. And rebase the branch

lirossarvet:
My personal pain point can be resolved with this by doing:

```ruby
# The below enables us to still rely on how the Beeline hooks into ActiveSupport 
# notfications while doing something other than calling `to_s` on every value
# in the payload Hash. For ActiveJob specifically, this means instead of setting
# the `adapter` field to something like "<SidekiqAdapter#0x12124124>" we get
# "SidekiqAdapter". 
#
# Similarly, in our sql.active_record custom listener, we can get specific 
# attributes of the connection into our Honeycomb event rather than simply
# stringifying the object (which doesn't have a friendly `to_s` call). 
# Additionally, you could use this method and simply not add the `binds` field
# to later scrub it with a presend_hook, as strongly suggested in the Ruby 
# beeline documentation, freeing up valuable space in a hook that runs for all
# events.

  # ...
  config.on_notification_event("perform.active_job") do |name, span, payload|
    span.add_field("#{name}.adapter", payload["adapter"].class)

    job = payload["job"]
    span.add_field("#{name}.job_class", job.class)

    span.add_field("#{name}.provider_id", job.provider_job_id)
    span.add_field("#{name}.scheduled_at", job.scheduled_at)
  end

  config.on_notification_event("sql.active_record") do |span, name, payload|
    span.add_field("#{name}.sql", payload["sql"])
    span.add_field("#{name}.name", payload["name"])
    span.add_field("#{name}.statement_name", payload["statement_name"])

    # Rails < 6 sends just connection_id, >= 6 sends an entire connection object
    if payload.key? "connection"
      span.add_field("#{name}.adapter", payload["connection"].adapter_name)
      span.add_field("#{name}.connection_id", payload["connection"].object_id)
    else
      span.add_field("#{name}.connection_id", payload["connection_id"])
    end
  end
end
```

ajvondrak:
> My personal pain point can be resolved with this by doing:

@lirossarvet This is beside the point, but are your args out of order here? :eyes: 😅 
```ruby
  config.on_notification_event("sql.active_record") do |span, name, payload|
```

lirossarvet:
> This is beside the point, but are your args out of order here? 

@ajvondrak why yes, yes they were. Fixed it.

robbkidd (approved):

robbkidd (approved):
I dig it.

Replacing default handler with a custom handler.

PR #153, opened 2021-06-17
Update Rails middleware to get status code even on raised error

lirossarvet:
Relates to #123! :D 

This came about, yet again, from a Pollinators conversation that inspired me to rethink this. Creating an overrideable `call_with_hook` allows Rails to do something _special_ to pull the status code and add package (controllers, routers, etc...) info even on error then bubble the error back up to the rack middleware to handle it as it always has been.


lib/honeycomb/integrations/rails.rb
```diff
@@ -89,6 +89,18 @@ class Middleware
...
+      def call_with_hook(env, span, &add_field)
+        super
+      rescue => e
+        wrapped = ActionDispatch::ExceptionWrapper.new(nil, e)
+
+        span.add_field "response.status_code", wrapped.status_code
+
+        raise e
+      ensure
+        add_package_information(env, &add_field)
```
martin308, line 102: maybe this should be a block passed into `call_with_hook`? Then we won't need to replicate the remainder of the above method

martin308 (commented):
A couple of lint errors to fix up but looks pretty straight forward to resolve. 

One suggestion on the `call_with_hook` rails implementation.

This seems like a reasonable solution to both backwards compatibility and doing the right thing though. Assuming all the tests pass.

Will leave the final stamp to someone on @honeycombio/integrations-team

lib/honeycomb/integrations/rails.rb
lirossarvet: 🤔 Hm. I'm not entirely sure I follow, can you elaborate a little?

martin308: presumably this ensure block is so that in the event of an exception that this information that is usually added on line 53 of the rack middleware will still be added? If so maybe `call_with_hook` should take a block and call that instead? Small example below

```
status, headers, body = call_with_hook(env, span) do
  add_package_information(env, &add_field)
  ... other things
end


def call_with_hook(env, span)
  super
  rescue => e
  raise e
  ensure
    yield
  end
end
```

lirossarvet: Ah okay, I think I follow now. The `ensure` block is there because adding that info seemed useful regardless of errors being raised. I think doing this in the `Rack` middleware may accomplish the same thing in a more generalized way...

```ruby
begin
  status, headers, body = call_with_hook(env, span, &add_field)
  span.add_field("response.status_code", status)
  span.add_field("response.content_type", headers["Content-Type"])
ensure
  add_package_information(env, &add_field)
  extract_user_information(env, &add_field)
end
```

This avoid the confusion of a thing like `call_with_hook` possibly taking two blocks and/or getting a bit convoluted about when the block is run. The above could also be added to the `Rack#call_with_hook` method rather than adding to `Rack#call` if that seems more fitting. I personally prefer it in `Rack#call` as it's more apparent what's being done. I wouldn't want the base `Rack#call_with_hook` to do anything so keeping those method calls in the main body feels clearer.

martin308: @lirossarvet even better!

lirossarvet: https://github.com/honeycombio/beeline-ruby/pull/153/commits/c512aa1d41ce56d6c6023d5775419865dc9d3eef @martin308 here ya go!

lirossarvet: Tested the `rails_52.gemfile` and `rails_6.gemfile` and they both passed locally at least. Not exhaustive but promising!

martin308: @lirossarvet great! If you fix the couple o lint errors then the CI should be able to run all the tests

lirossarvet: https://github.com/honeycombio/beeline-ruby/pull/153/commits/ec25a14fa6fb7ba1792e29e2f8bd8fe5a7ee4aae

Done! @martin308 thanks so much for the feedback

lirossarvet:
Okay I'm pretty happy tests passed! @martin308 I know the integrations team specifically needs to interview this, but thanks again for your feedback. Let me know if there's anything else I need to do.

lib/honeycomb/integrations/rails.rb
```diff
@@ -89,6 +89,16 @@ class Middleware
       include Rack
       include Warden
       include Rails
+
+      def call_with_hook(env, span, &_add_field)
+        super
+      rescue StandardError => e
+        wrapped = ActionDispatch::ExceptionWrapper.new(nil, e)
+
+        span.add_field "response.status_code", wrapped.status_code
```
lirossarvet, line 98: One final note on this, maybe: there's a *chance* this status code is wrong since a middleware higher in the stack may change it. However, that's always been true given the tracing middleware isn't the first in/last out one.

paulosman (approved):
Thank you!

PR #154, opened 2021-06-23
make Rails spec consistent with Honeycomb Railtie initialization

robbkidd:
With #153 merged, this addresses the TODO comments left from adding tests for the events generated from error responses in #132.

Also adds a few more tests for the `error` and `error_details` fields that are the candy-like center of this effort to capture the details of errors in Rails.

### DIFF
--- a/spec/honeycomb/integrations/rails_spec.rb
+++ b/spec/honeycomb/integrations/rails_spec.rb
@@ -37,8 +37,6 @@ if defined?(Honeycomb::Rails)
     let(:client) { Honeycomb::Client.new(configuration: configuration) }
     let(:app) do
       Class.new(Rails::Application).tap do |app|
-        app.config.logger = Logger.new(STDERR)
-        app.config.log_level = :fatal
         app.config.eager_load = false
         app.config.secret_key_base = "3b7cd727ee24e8444053437c36cc66c4"
         app.config.respond_to?(:hosts) && app.config.hosts << "example.org"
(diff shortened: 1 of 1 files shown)


spec/honeycomb/integrations/rails_spec.rb
```diff
@@ -109,38 +108,10 @@ def hello_error
...
-
       before do
         get "/hello_error?honey=bee"
       end
 
-      # TODO: remove these lets when we add shared_examples
-      # because they're declared in
-      let(:event_data) { libhoney_client.events.map(&:data) }
-      let(:event) { event_data.first }
-
```
martin308, line 143: the comment implies this should be replaced with the shared examples but I don't see them being added here

robbkidd: It does imply that! And when I went to add [the shared examples](https://github.com/honeycombio/beeline-ruby/blob/03715e86ec9291c1a27ce0fe0e26d25bacdfde1f/spec/honeycomb/integrations/rails_spec.rb#L72-L80), I found [they were already there](https://github.com/honeycombio/beeline-ruby/blob/03715e86ec9291c1a27ce0fe0e26d25bacdfde1f/spec/honeycomb/integrations/rails_spec.rb#L139-L160).

martin308 (approved):
Looks good. Some tests are failing though as it looks like different rails versions handle this error differently. May need to alter the expectations based on rails versions like we do down here https://github.com/honeycombio/beeline-ruby/blob/main/spec/honeycomb/integrations/rails_spec.rb#L279

robbkidd:
> Some tests are failing though as it looks like different rails versions handle this error differently.

🧐  Indeed. Investigating. Thank you for the pointer to setting different expectations!

robbkidd:
@martin308 You already had a "Rails 5 or greater" in that `describe` context; I just had to open my eye to see it.

PR #155, opened 2021-06-23
CI improvements

robbkidd:
- [x] make the CI job names easier to read
- [x] only install bundler once
-  ~update to using the ["next gen" images](https://circleci.com/docs/2.0/circleci-images/#next-generation-convenience-images) provided by CircleCI~
    - Can't do this easily because Circle doesn't produce next gen images for Ruby 2.2 and 2.3 because of course not, those are old. Using one type of image for old Ruby and another type for newer Ruby would make the matrix complicated, so punting on this as "not low-hanging fruit."
- [x] resolve deprecation warning when using `bundle install --path vendor/bunder`
- [x] add a JUnit formatter to RSpec and have CircleCI store results

Bonus:

- [x] Rails 6.1!


martin308 (approved):
Amazing! The only thing I could think of to make it better would be if the job names included ruby in there to be more explicit like this

- test-faraday_0-ruby_2.7 - Success

robbkidd:
> [Could] the job names included ruby[?]

I had the same thought! I only just figured out how to set the job names while preserving the matrix'd parameters. Which order do you like more?

1. `test-faraday_0-ruby_2.7`
2. `test-ruby_2.7-faraday_0`

robbkidd:
I went with №1 because that's the least change. We can fiddle with display more in the future if we like.

:shipit: 🚀 🍰

PR #160, opened 2021-07-08
Allow backtrace to be sent with errors

lirossarvet:
I often find myself looking for errors and finding something like:

```
NoMethodError: No method `split' on <String#0x00012121212>
```

This is decidedly unhelpful for figuring out where exactly this happened, especially if the specific code isn't instrumented because normally it's trivial. This means the error only bubbles up and exists on the next-nearest parent or, even worse, the root span. By allowing users to configure and provide a limited set of the backtrace, they can get at least the chance for context without finding the error'ed event in Honeycomb then trying to find the same error in their logs to get the backtrace of the location.

While the limited subset _may still_ mean they have to do that, this at least opens the door to providing *just* enough context to find where an error is coming from.


lirossarvet:
Note: this came about from a Slack question and conversation with @robbkidd in Pollinators

lib/honeycomb/client.rb
```diff
@@ -38,6 +38,7 @@ def initialize(configuration:)
         parser_hook: configuration.http_trace_parser_hook,
         propagation_hook: configuration.http_trace_propagation_hook,
       }
+      @error_backtrace_limit = configuration.error_backtrace_limit.to_i
```
lirossarvet, line 41: While the Configuration also does what it can to ensure this is an `Integer` at load/configuration time, it seemed good to just try and convert this here on the safe side.

lib/honeycomb/client.rb
```diff
@@ -47,16 +48,7 @@ def initialize(configuration:)
...
-        Trace.new(serialized_trace: serialized_trace,
-                  builder: libhoney.builder,
-                  context: context,
-                  **@additional_trace_options)
-      else
-        context.current_span.create_child
-      end
-
-      current_span = context.current_span
+      current_span = get_current_span(serialized_trace: serialized_trace)
```
lirossarvet, line 51: Rubocop said this method was too long now, so I just shuffled this "create a Trace if there isn't one, otherwise make a child from the current span" logic into a private method. Let me know if disabling Rubocop here would be preferable.

lib/honeycomb/configuration.rb
```diff
@@ -12,19 +12,25 @@ class Configuration
...
+      @error_backtrace_limit = 0
       @client = nil
     end
 
     def service_name
       @service_name || dataset
     end
 
+    def error_backtrace_limit=(val)
+      @error_backtrace_limit = Integer(val)
```
lirossarvet, line 31: Get a `TypeError` at config time if you set a non-numerical value here rather than waiting until later to either not get your values or something.

lib/honeycomb/client.rb
```diff
@@ -71,6 +63,19 @@ def start_span(name:, serialized_trace: nil, **fields)
       rescue StandardError => e
         current_span.add_field("error", e.class.name)
         current_span.add_field("error_detail", e.message)
+
+        if error_backtrace_limit > 0
+          # Honeycomb by default escapes things like brackets and shows the
+          # unicode sequence. So <main> becomes \u003cmain\u003e. Given the
+          # confusion that could cause in readability, we replace < and > with
+          # a single quote.
+          backtrace = e.backtrace[0...error_backtrace_limit]
```
MikeGoldsmith, line 72: Is it possible to know how many levels the back trace has? If we know the limit is >= number of levels, we will have a complete backtrace not a partial.

MikeGoldsmith (commented):
I think this look great, thank you for the contribution @lirossarvet.

I have one query that would allow us to indicate if we've captured a whole or partial backtrace.

lib/honeycomb/client.rb
robbkidd: It is possible. `e.backtrace` is an array, so `e.backtrace.length` or `.size` would tell us how many levels there are. Are you picturing changing the name of the field being added–`error_backtrace_partial` or `error_backtrace`–based on whether the backtrace length is less than or equal to the limit?

lirossarvet: I'd be more inclined to leave this as the same name no matter what (and maybe just `error_backtrace`) but introduce two additional columns: `error_backtrace_limit` and `error_backtrace_length`. With those two, someone could make a derived column of, like, `error_backtrace_is_complete` or something by doing a quick comparison.

But maybe adding 3 new columns with 1 PR is a bit much.

MikeGoldsmith: @robbkidd and I have just discussed and came to the same conclusion that event fields for the backtrace, length and limit would be better as it would provide consistent field names.

We're like your suggestions of:
```
error_backtrace
error_backtrace_length
error_backtrace_limit
```

If you're happy to update to add them that would great, we can add them.

lirossarvet: I did `error_backtrace_total_length` but am happy to undo that little bit. 

https://github.com/honeycombio/beeline-ruby/pull/160/commits/7d3eb6f187783664db2b8b16f591342162bed203

MikeGoldsmith (approved):
Looks great, thanks for the updates @lirossarvet 👍🏻

lib/honeycomb/client.rb
```diff
@@ -69,8 +61,8 @@ def start_span(name:, serialized_trace: nil, **fields)
       begin
         yield current_span
       rescue StandardError => e
-        current_span.add_field("error", e.class.name)
-        current_span.add_field("error_detail", e.message)
+        add_exception_data(current_span, e)
```
robbkidd, line 64: I see what you did here ... and I like it.

lirossarvet: You can, once again, think Rubocop for making me multi-line these function calls them yelling at me about the method being too long -.-

robbkidd: And you came up with an elegant solution to reconcile the "you should call this exception `e`" cop fighting with our desire for a longer variable name so that the indentation of the method chaining doesn't make our eyes itch. Thank you for the nagging, Rubocop.

lirossarvet:
For anybody who comes looking, this can be setup by setting a new config option `error_backtrace_limit` to some positive integer. I would suggest something like 5 or 7. Most Rails apps are going to have a monstrous backtrace that may cause your event (or an entire batch) to be rejected if you send too much information in it. The goal was to provide enough context you could find where exactly in your span this came from since a span may wrap 3-4 layers of stack flow.

PR #166, opened 2021-08-27
normalize ActiveSupport::Notification exceptions into error / error_d…

leviwilson:
There was a discussion in the [`#sdk-ruby`](https://honeycombpollinators.slack.com/archives/CJR134U2F/p1630081866001300)  channel regarding `ActiveSupport::Notification` payloads normalizing the `payload[:exception]` similarly to the way that other integrations in the beeline by setting `"error"` and `"error_detail"` on the span.

Making this consistent will assist in querying traces that might contain exceptions.

* handle when a `payload[:exception]` exists by adding:
  * `error` - the exception class
  * `error_detail` - the exception message

One decision that I was not clear about was what to do about the existing information on the payload. This means that if you have an event such at `"muh.custom.thing"` and an error occurrs, you will have this:

```json
{
  "muh.custom.thing.exception": ["StandardError", "I tried, but I have failed"],
  "error": "StandardError",
  "error_detail": "I tried, but I have failed"
}
```

My thoughts regarding this were to leave it untouched, so as to not break any expectations.

### DIFF
--- a/lib/honeycomb/integrations/active_support.rb
+++ b/lib/honeycomb/integrations/active_support.rb
@@ -50,6 +50,13 @@ module Honeycomb
             # Make ActionController::Parameters parseable by libhoney.
             value = value.to_unsafe_hash if value.respond_to?(:to_unsafe_hash)
             span.add_field("#{name}.#{key}", value)
+
+            error, error_detail = payload[:exception]
+
+            if error
+              span.add_field("error", error)
+              span.add_field("error_detail", error_detail)
+            end
           end
         end
       end
--- a/lib/honeycomb/integrations/active_support.rb
+++ b/lib/honeycomb/integrations/active_support.rb
@@ -52,11 +52,12 @@ module Honeycomb
             span.add_field("#{name}.#{key}", value)
           end
 
-          exception_object = payload[:exception_object]
-          next unless exception_object
-
-          span.add_field("error", exception_object.class.to_s)
-          span.add_field("error_detail", exception_object.message)
+          # If the notification event has recorded an exception, add the
+          # Beeline's usual error fields to the span.
+          if payload[:exception_object]
+            span.add_field("error", payload[:exception_object].class.to_s)
+            span.add_field("error_detail", payload[:exception_object].message)
+          end
         end
       end
     end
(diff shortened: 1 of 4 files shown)


martin308 (commented):
Will leave the approval up to @robbkidd. This looks good to me though as an idea. 

My only suggestion here would be to add some negative test cases around different values being in the `exception` key. It looks like this code would handle them but would be nice to codify that behaviour.

lib/honeycomb/integrations/active_support.rb
```diff
@@ -50,6 +50,13 @@ def default_handler
             # Make ActionController::Parameters parseable by libhoney.
             value = value.to_unsafe_hash if value.respond_to?(:to_unsafe_hash)
             span.add_field("#{name}.#{key}", value)
+
+            error, error_detail = payload[:exception]
```
robbkidd, line 54: Aww yiss, array expansion. I think, though, that since you're grabbing the key off `payload` this ought to happen outside the `payload.each do … end` loop. The end result is correct, but this is extra work inside the loop.

spec/honeycomb/integrations/active_support_spec.rb
```diff
@@ -173,6 +173,26 @@
         expect(this_event.data["honeycomb.nonspecific_event.honeycomb"]).to eq 1
       end
 
+      it "passes along exception information" do
```
robbkidd, line 176: This was a dope test. 👍🏻 

It helped me get my brain around adding a whole new case with nothing very custom for the default handler. I'd like to push that test case to this PR's branch, if you are OK with that and have the maintainers can update this PR enabled.

robbkidd (changes requested):
A couple of comments inline with the code.

* The assignment and `add_field`ing needs to move out of the `payload.each` loop.
* Consider reaching for the `exception_object` instead of `exception` on `payload`.

lib/honeycomb/integrations/active_support.rb
leviwilson: Derp...not sure why I didn't see that :+1:

spec/honeycomb/integrations/active_support_spec.rb
leviwilson: @robbkidd go for it (and I did choose that option)

robbkidd (approved):
Okay. Let's ride.

robbkidd:
Thank you for your contribution, @leviwilson!

PR #172, opened 2021-09-14
remove CodeCov from test suite

### LINKED ISSUE #168: Address CodeCov gem deprecation and the errors in the nightly.
The nightlies are failing because of CodeCov's API being upset that our very wide test matrix submits coverage reports for the same commit (main's current) in quick succession.

1. We don't really need coverage tracking reported for the nightly build. We ought to turn off reporting in that workflow.
2. [CodeCov has switched to using an external binary to report coverage instead of language-specific libraries](https://about.codecov.io/blog/codecov-uploader-deprecation-plan/). If we're going to keep CodeCov, we ought to switch to that.

### LINKED ISSUE #170: Do we need the Gemfile.lock checked in?
I think maybe not. Removing it from version control could reduce the number of PRs dependabot has to open to keep locked version pins up-to-date.

robbkidd:
Remove CodeCov. (Closes #168.) We're not tracking coverage overtime as much as we used to. SimpleCov is still available to help determine if there is sufficient coverage of a change.

Also remove the main Gemfile.lock (Closes #170.) to allow gem versions to float to latest resolveable by gemspec and Ruby version constraints.

tenor-23748719


robbkidd:
tenor-23748719

spec/spec_helper.rb
```diff
@@ -1,19 +1,15 @@
...
 require "simplecov-console"
-require "codecov" if SUPPORT_CODECOV
 require "webmock/rspec"
 require "pry"
 
 WebMock.disable_net_connect!
 
 Dir["./spec/support/**/*.rb"].sort.each { |f| require f }
 SimpleCov.formatter = SimpleCov::Formatter::Console
-SimpleCov.formatter = SimpleCov::Formatter::Codecov if SUPPORT_CODECOV
```
vreynolds, line 16: why remove this line only?

robbkidd: It's the line that configures the coverage generating tool (SimpleCov) to use the formatter that sends the report to CodeCov. Without this line, the previous line remains in effect: report current code coverage to STDOUT.

vreynolds (approved):
bbb

spec/spec_helper.rb
robbkidd: Compare the console output end in [a test from this PR](https://app.circleci.com/pipelines/github/honeycombio/beeline-ruby/999/workflows/52f6a9f5-ebce-4842-92ef-884c7d8e66aa/jobs/71185) … 

Screen Shot 2021-09-14 at 6 26 27 PM


 to [a test from the past](https://app.circleci.com/pipelines/github/honeycombio/beeline-ruby/995/workflows/717907b7-499c-415f-9da9-ae267c83215a/jobs/70833).

Screen Shot 2021-09-14 at 6 27 12 PM

PR #173, opened 2021-09-17
gather up the matrix'd code coverage and save a report

robbkidd:

- We removed CodeCov which was what had been aggregating the code coverage reports across the matrix of test jobs. Without aggregating the results, each report looks like things are left without coverage.

- Cache the coverage report for a particular job in a job-named directory, cache those results in the CI workspace, collate the results once all the test jobs are complete, then store the report in the over-all job's artifacts (coverage directory, though, not the artifacts directory).


MikeGoldsmith (approved):

robbkidd:
CI-only changes, so no version bump or release necessary.

PR #178, opened 2021-10-20
fix: new redis gem version broke tests

### LINKED ISSUE #177: beeline-ruby redis tests failing on main
When preparing next release on #146 , realized build is failing redis tests. This needs to get sorted out before a release can go out

vreynolds:

- redis gem update includes additional READONLY commands being executed in cluster mode, which results in additional events generated by the beeline
- fixes #177 

- decided to keep the tests specific and create a new IF branch for the newer versions of the gem

### DIFF
--- a/spec/honeycomb/integrations/redis_spec.rb
+++ b/spec/honeycomb/integrations/redis_spec.rb
@@ -2054,7 +2054,13 @@ if defined?(Honeycomb::Redis)
 
         let(:redis) do
           allow(connection).to receive(:read).and_return(slots, nodes, commands)
-          Redis.new(driver: driver, cluster: cluster, replica: true).tap do
+          Redis.new(
+            driver: driver, cluster: cluster, replica: true,
+          ).tap do |redis|
+            # https://github.com/redis/redis-rb/issues/1017
+            # establishes connections to all nodes in the cluster,
+            # otherwise extra READONLY commands are executed on replicas
+            redis.auth
             libhoney_client.events.clear
           end
         end
--- a/spec/honeycomb/integrations/redis_spec.rb
+++ b/spec/honeycomb/integrations/redis_spec.rb
@@ -2057,9 +2057,9 @@ if defined?(Honeycomb::Redis)
           Redis.new(
             driver: driver, cluster: cluster, replica: true,
           ).tap do |redis|
-            # https://github.com/redis/redis-rb/issues/1017
-            # establishes connections to all nodes in the cluster,
+            # Establishes connections to all nodes in the cluster,
             # otherwise extra READONLY commands are executed on replicas
+            # due to https://github.com/redis/redis-rb/issues/1017
             if VERSION > Gem::Version.new("4.1.1")
               redis.auth
             else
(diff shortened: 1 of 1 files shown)


vreynolds:
Holy shirt balls. 🤯 You need a decompression chamber after that dive, @ajvondrak ? 🤣

Good catch! I see now that the change was introduced in v4.4.0 -- I like the idea of matching our test cases to that for consistency.

ajvondrak:
Alternatively, we could prevent the `READONLY` spans from appearing in v4.5 by forcing a connection to every node in the cluster prior to the test assertions:

```diff
diff --git a/spec/honeycomb/integrations/redis_spec.rb b/spec/honeycomb/integrations/redis_spec.rb
index e834585..256afed 100644
--- a/spec/honeycomb/integrations/redis_spec.rb
+++ b/spec/honeycomb/integrations/redis_spec.rb
@@ -2054,7 +2054,8 @@

         let(:redis) do
           allow(connection).to receive(:read).and_return(slots, nodes, commands)
-          Redis.new(driver: driver, cluster: cluster, replica: true).tap do
+          Redis.new(driver: driver, cluster: cluster, replica: true).tap do |redis|
+            redis.auth # establishes connections to all nodes in the cluster
             libhoney_client.events.clear
           end
         end
```
This makes the old tests pass with v4.5.1 on my machine. Avoids the problem at all, and is a little less hacky than my initial crack at dropping the spans with the sampling hook. 😅 Unless we want to test the scenario where the connection happens in the middle of the cluster commands (thus triggering a `READONLY`), in which case we should keep pulling this thread with the new assertions you've written.

spec/honeycomb/integrations/redis_spec.rb
```diff
@@ -2054,7 +2054,17 @@
...
+          Redis.new(
+            driver: driver, cluster: cluster, replica: true,
+          ).tap do |redis|
+            # https://github.com/redis/redis-rb/issues/1017
+            # establishes connections to all nodes in the cluster,
+            # otherwise extra READONLY commands are executed on replicas
+            if VERSION > Gem::Version.new("4.1.1")
+              redis.auth
+            else
+              redis.auth("")
```
ajvondrak, line 2066: Oh, hm. 🤔 I assume it makes sense to pass a string (the password for the `AUTH` command) on all versions of Redis, rather than switch off of v4.1.1. That's what the other tests do: https://github.com/honeycombio/beeline-ruby/blob/adf2e8c3c27e308a94be8bd8d285b7ec92ed40e1/spec/honeycomb/integrations/redis_spec.rb#L185-L185 I just neglected to include it in the diff I commented with before. 😅

spec/honeycomb/integrations/redis_spec.rb
```diff
@@ -2054,7 +2054,17 @@
 
         let(:redis) do
           allow(connection).to receive(:read).and_return(slots, nodes, commands)
-          Redis.new(driver: driver, cluster: cluster, replica: true).tap do
+          Redis.new(
+            driver: driver, cluster: cluster, replica: true,
+          ).tap do |redis|
+            # https://github.com/redis/redis-rb/issues/1017
+            # establishes connections to all nodes in the cluster,
+            # otherwise extra READONLY commands are executed on replicas
```
ajvondrak, line 2062: Nit: my head parsed this comment a little strangely. I thought it was saying that the code in the GitHub issue established a connection. Maybe reverse the phrasing?
```suggestion
            # Establishes connections to all nodes in the cluster,
            # otherwise extra READONLY commands are executed on replicas
            # due to https://github.com/redis/redis-rb/issues/1017
```

JamieDanielson (approved):
🔑

PR #191, opened 2022-02-02
Remove the upperbound on libhoney version

MikeGoldsmith:
We have a pessimistic version guard on libhoney that prevents us from using 2.x. We should remove it.

- Removes dependency guard of libhoney

### DIFF
--- a/honeycomb-beeline.gemspec
+++ b/honeycomb-beeline.gemspec
@@ -37,7 +37,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_dependency "libhoney", ">= 1.14.2", "~> 1.14"
+  spec.add_dependency "libhoney", ">= 1.14.2"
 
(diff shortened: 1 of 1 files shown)


robbkidd (approved):
FREEDOM

honeycomb-beeline.gemspec
```diff
@@ -37,7 +37,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_dependency "libhoney", ">= 1.14.2", "~> 1.14"
```
martin308, line 40: you might want a `< 3` in here to prevent using potentially breaking changes in the future?

robbkidd: @MikeGoldsmith and I talked about that and see that point. In the end though, we concluded that we'd rather put a guard in later if necessary than require a Beeline upgrade to use a new libhoney. Particularly in cases like this, where the major version bump for libhoney was not a change to the public API, but Ruby version drop-support.

PR #195, opened 2022-03-17
wip e&s update...

### LINKED ISSUE #193: Update for E&S
**Is your feature request related to a problem? Please describe.**

Environments & Services are coming to Honeycomb. The beeline should be updated to support sending data to environments while maintaining backward compatibility. To achieve this, we will set dataset from service name, and provide appropriate warnings depending on environment vs. non-environment teams.

**Describe the solution you'd like**

Tracing scenario for environment-aware teams:

* Dataset should no longer propagate across services for new environment key.
* `ServiceName` "required"†. If unset, will warn and set `service_name` AND `service.name` to `unknown_service:<process>` or `unknown_service:language`. Service name is used to populate dataset name; if `unknown_service.*` then truncate to`unknown_service` for dataset name. `service.name` field will be added to the already added `service_name` for closer consistency with OTel.
* `WriteKey` required. If unset, will warn.
* `Dataset` **ignored**. If unset, no warning (because no longer used). If set, will warn to clarify that it will be ignored in favor of service name.

Tracing scenario for legacy (non-environment-aware) teams:

* `ServiceName` "required"†. If unset, will warn.
* `WriteKey` required. If unset, will warn.
* `Dataset` required. If unset, will warn and default to `beeline-<language>`.

† While not technically required, service name is highly encouraged to avoid `unknown_service` and to instead properly describe the data being sent to Honeycomb (and the data being viewed in Honeycomb UI).

This will be addressed in a separate issue/PR: Set [spanKind value based on OTel spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#spankind) for consistency across beelines and OTel.

### Notes from decision around defaulting when no service name is provided:
*Given* a beeline configured with a v2 api key and without a service name
*When* it sends HNY events
*Then* it should use `unknown_service` as the dataset
*And* it should use `unknown_service:process-name` as service_name

When given a legacy key, the behavior remains the same as today.

[Notes on trimming whitespace](https://honeycomb.quip.com/qdrvA6a9AHEI/notes-trimming-whitespace-ingest)

JamieDanielson:

- closes #193 

-

### DIFF
--- a/lib/honeycomb/client.rb
+++ b/lib/honeycomb/client.rb
@@ -30,6 +30,7 @@ module Honeycomb
 
       # maybe make `service_name` a required parameter
       @libhoney.add_field "service_name", configuration.service_name
+      @libhoney.add_field "service.name", configuration.service_name
       @context = Context.new
 
       @additional_trace_options = {
--- a/lib/honeycomb/configuration.rb
+++ b/lib/honeycomb/configuration.rb
@@ -14,6 +14,11 @@ module Honeycomb
     attr_writer :service_name, :client, :host_name
     attr_reader :error_backtrace_limit
 
+    default_dataset_classic = "rails"
+    default_dataset = "unknown_service"
+    # TODO get process for unknown_service name?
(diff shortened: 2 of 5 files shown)


JamieDanielson:
🙈

PR #196, opened 2022-03-18
Add Environment and Services support

### LINKED ISSUE #193: Update for E&S

MikeGoldsmith:

- closes #193 

If using an environment-aware key:
- dataset = serviceName
- add field of `service.name` in addition to `service_name` to be more consistent with otel
- use default service name of `unknown_service:<processname>` or `unknown_service:ruby` if process name is unavailable and no service name provided
- don't propagate dataset by default unless legacy key (to prevent overwriting of service name)
- warn on missing service name and api key
- warn on presence of dataset, noting data will be sent to service name
- if dataset derived from service name starts with `unknown_service*`, truncate to `unknown_service` for dataset
- if dataset derived from service name has extra whitespace, trim whitespace and console warn on diff

warning provided if dataset is set in config when using environment key: 
`dataset will be ignored, sending data to message-rb`
`service_name is unknown, will set to unknown_service:rackup`
`dataset will be ignored, sending data to unknown_service:rackup`

If using a classic key:
- dataset is dataset in config, no default set
- add field of `service.name` in addition to `service_name` to be more consistent with otel
- use default service name of `unknown_service:<processname>` or `unknown_service:go` if process name is unavailable and no service name provided
- dataset propagates by default
- warn on missing service name, api key, dataset

warning on missing env vars with classic key:
`empty service_name option`
`empty dataset option`
`Libhoney::Client: no dataset configured, disabling sending events`

### DIFF
--- a/lib/generators/honeycomb/honeycomb_generator.rb
+++ b/lib/generators/honeycomb/honeycomb_generator.rb
@@ -12,6 +12,7 @@ class HoneycombGenerator < Rails::Generators::Base
   argument :write_key, required: true, desc: "required"
 
   class_option :dataset, type: :string, default: "rails"
+  class_option :service_name, type: :string, default: "rails"
 
   gem "honeycomb-beeline"
 
@@ -22,7 +23,13 @@ class HoneycombGenerator < Rails::Generators::Base
       <<-RUBY.strip_heredoc
         Honeycomb.configure do |config|
           config.write_key = #{write_key.inspect}
-          config.dataset = #{options['dataset'].inspect}
+          def classic?
+            config.write_key.nil || config.write_key.length == 32
+          end
+          if classic?
+            config.dataset = #{options['dataset'].inspect}
+          else
+            config.dataset = #{options['service_name'].inspect}
           config.presend_hook do |fields|
             if fields["name"] == "redis" && fields.has_key?("redis.command")
               # remove potential PII from the redis command
--- a/lib/generators/honeycomb/honeycomb_generator.rb
+++ b/lib/generators/honeycomb/honeycomb_generator.rb
@@ -23,10 +23,7 @@ class HoneycombGenerator < Rails::Generators::Base
       <<-RUBY.strip_heredoc
         Honeycomb.configure do |config|
           config.write_key = #{write_key.inspect}
-          # def classic?
-          # config.write_key.nil? || config.write_key.length == 32
           classic = config.write_key.nil? || config.write_key.length == 32
-          # end
           if classic
             config.dataset = #{options['dataset'].inspect}
           else
--- a/lib/generators/honeycomb/honeycomb_generator.rb
+++ b/lib/generators/honeycomb/honeycomb_generator.rb
@@ -23,10 +23,11 @@ class HoneycombGenerator < Rails::Generators::Base
       <<-RUBY.strip_heredoc
         Honeycomb.configure do |config|
           config.write_key = #{write_key.inspect}
-          def classic?
-            config.write_key.nil? || config.write_key.length == 32
-          end
-          if classic?
+          # def classic?
+          # config.write_key.nil? || config.write_key.length == 32
+          classic = config.write_key.nil? || config.write_key.length == 32
+          # end
+          if classic
             config.dataset = #{options['dataset'].inspect}
           else
             config.dataset = #{options['service_name'].inspect}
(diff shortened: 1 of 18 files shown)


spec/generators/honeycomb/honeycomb_generator_spec.rb
```diff
@@ -6,7 +6,8 @@
   RSpec.describe HoneycombGenerator do
     describe "simple execution" do
       let(:name) { "honeycomb" }
-      let(:write_key) { "generator_write_key" }
+      # let(:write_key) { "generator_write_key" }
```
vreynolds, line 9:
```suggestion
```

spec/honeycomb/configuration_spec.rb
```diff
@@ -2,15 +2,12 @@
 
 RSpec.describe Honeycomb::Configuration do
   let(:configuration) { Honeycomb::Configuration.new }
-  let(:dataset_name) { "dataset" }
-  let(:service_name) { "service_name" }
-  let(:write_key) { "service_name" }
+  let(:write_key) { "d68f9ed1e96432ac1a3380" }
```
vreynolds, line 5: Is this a made-up key?

vreynolds (approved):
Looks good, just a couple small notes

spec/honeycomb/configuration_spec.rb
JamieDanielson: AFAIK it was randomly generated by @MikeGoldsmith ... I believe I asked the same question 😅

spec/generators/honeycomb/honeycomb_generator_spec.rb
JamieDanielson: oops good catch

lib/generators/honeycomb/honeycomb_generator.rb
```diff
@@ -22,7 +23,12 @@ def create_initializer_file
       <<-RUBY.strip_heredoc
         Honeycomb.configure do |config|
           config.write_key = #{write_key.inspect}
-          config.dataset = #{options['dataset'].inspect}
+          classic = config.write_key.nil? || config.write_key.length == 32
+          if classic
+            config.dataset = #{options['dataset'].inspect}
+          else
+            config.dataset = #{options['service_name'].inspect}
+          end
```
robbkidd, line 31: This is code that gets added to the end-users code base the first time they install the beeline into a rails app. We could consider making this generator only do the not-classic behavior under the assumption that anyone adding the beeline to their app will be a new customer with an environment.

robbkidd (approved):
:shipit:

PR #200, opened 2022-04-16
Let Rake spans be disabled programmatically

ajvondrak:

When running a Rake task, the Honeycomb integration typically creates the root node of the trace (since the task is the entry point into the program). For short-lived tasks like `rake db:migrate`, this works great. However, tasks might be long-lived. For example, `rake resque:work` [runs forever](https://github.com/resque/resque/blob/e068909cc0cbf1afa2fa4cb33be2809a397ce9d3/README.markdown#running-workers), dequeueing background jobs on a loop. Ideally, we'd like a separate trace for each job that gets dequeued. But with the Rake integration enabled, all the jobs' traces point back up at the auto-generated root span, which will be missing until the worker process finally dies. This has been discussed before in Pollinators: https://honeycombpollinators.slack.com/archives/CJR134U2F/p1643989199035879

As-is, the only way to disable the Rake integration is to remove it from the `HONEYCOMB_INTEGRATIONS` env var. There's a bug in the implementation of `Rake::Application#honeycomb_client` such that setting it to `nil` doesn't disable the Rake tracing (same sort of bug as in #60). Moreover, we can't pick & choose which Rake tasks generate a span. So if we disable the Rake integration altogether, `rake resque:work` might look good, but something like `rake db:migrate` will just generate a bunch of orphans that can't be tied back to a single parent.

This PR allows you to programmatically disable the Rake integration both on the application level and the individual task level.

- Fixed bug in `Rake::Application#honeycomb_client` so that it will honor when we set it to `nil`.
- Modified `Rake::Task#honeycomb_client` so that it can be customized similarly to the application-level client.
- Overhauled the specs, which had not been very extensive previously. You can still see the original test in the form of the `test:client:access` task.

With this, you can selectively disable individual Rake tasks' auto-generated spans by saying

```ruby
Rake::Task["resque:work"].honeycomb_client = nil
```

Alternatively, you could disable Rake at the application level and opt-in to the tracing at the task level:

```ruby
Rake.application.honeycomb_client = nil
Rake::Task["db:migrate"].honeycomb_client = Honeycomb.client
```

Other thoughts:
- I refrained from elevating this to a `Honeycomb::Configuration` option just to keep the change as simple as possible. It's perhaps not the most user-friendly, but it _is_ maximally flexible: use whatever arbitrary logic you want to toggle the client to `nil`.
- As common as the Resque use case is, I wonder if it might be worth adding something to the Rails generator. :thinking: But then, I haven't tested out which exact Resque spans to disable this way, yet.

### DIFF
--- a/lib/honeycomb-beeline.rb
+++ b/lib/honeycomb-beeline.rb
@@ -25,7 +25,7 @@ module Honeycomb
     extend Forwardable
     attr_reader :client
 
-    def_delegators :@client, :libhoney, :start_span, :add_field,
+    def_delegators :client, :libhoney, :start_span, :add_field,
                    :add_field_to_trace, :current_span, :current_trace,
                    :with_field, :with_trace_field
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -13,6 +13,7 @@ Metrics/BlockLength:
   Exclude:
     - honeycomb-beeline.gemspec
     - spec/**/*.rb
+    - spec/**/*.rake
 
 Metrics/AbcSize:
   Max: 50
--- a/lib/honeycomb/integrations/rake.rb
+++ b/lib/honeycomb/integrations/rake.rb
@@ -35,7 +35,9 @@ module Honeycomb
       attr_writer :honeycomb_client
 
       def honeycomb_client
-        @honeycomb_client || Honeycomb.client
+        return @honeycomb_client if defined?(@honeycomb_client)
+
+        Honeycomb.client
       end
     end
   end
--- a/lib/honeycomb/integrations/rake.rb
+++ b/lib/honeycomb/integrations/rake.rb
@@ -22,7 +22,11 @@ module Honeycomb
         end
       end
 
+      attr_writer :honeycomb_client
+
       def honeycomb_client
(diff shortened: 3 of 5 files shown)


lib/honeycomb-beeline.rb
```diff
@@ -25,7 +25,7 @@ class << self
     extend Forwardable
     attr_reader :client
 
-    def_delegators :@client, :libhoney, :start_span, :add_field,
+    def_delegators :client, :libhoney, :start_span, :add_field,
```
ajvondrak, line 28: This is sort of a subtle change. By going through `:client` instead of `:@client`, this allows us to use RSpec to stub `Honeycomb.client` like

```ruby
allow(Honeycomb).to receive(:client).and_return(...)
```

And then `Honeycomb.start_span` & company will go through the stubbed client.

This is important because there's no way (short of `instance_variable_set`) to reset the client after a `Honeycomb.configure`. Thus, a `Honeycomb.configure` in some stray `before` block could leak into subsequent tests, which might break things depending on the order they're run. It's probably worth going through the rest of the suite and cleaning up the few stray uses of `Honeycomb.configure`.

PR #201, opened 2022-04-21
Console: Load the file that exists

pusewicz:

- The `bin/console` script not working

- The `honeycomb/beeline` does not exist and this script does not work. Instead load the relative file from the source tree.


robbkidd:
`bin/console` is wholly a development environment tool, so this change does not require a version bump or a trigger for release.

Issue #207, opened 2022-09-13
Failing nightly build - Rack is unhappy


robbkidd:
**Versions**

- Ruby: all the versions in the test matrix
- Beeline: [main](https://github.com/honeycombio/beeline-ruby/tree/c82a29dd0c8bf75ddafe4bc05d0ee1161fe7b0b9) as of c82a29dd0c8bf75ddafe4bc05d0ee1161fe7b0b9

**Steps to reproduce**

1. See the Rack failures in [a sad nightly build](https://app.circleci.com/pipelines/github/honeycombio/beeline-ruby/1511/workflows/aeea2e4f-6a25-4122-a374-23bf37a37449)

**Additional context**

PR #208, opened 2022-09-16
wip - fix up broken nightly tests

JamieDanielson:
## Which problem is this PR solving?

- closes #207 

## Short description of the changes

-

PR #209, opened 2022-09-26
handle recent Rack v3 changes in tests & example

emilyashley:

- #207 

- Rack::Lobster moved to Rackup::Lobster in newest versions, we removed lobster.
- Build a Rack 2 and a Rack 3 and use in appraisals and CI
- conditionally catch which version of rack is being used and use appropriate code 
- Rack Session also moved rack 3 as well ( and longer secret is required) 
- replace ::Rack::VERSION with ::Rack.release


spec/honeycomb/integrations/rack_spec.rb
```diff
@@ -2,14 +2,26 @@
 
 if defined?(Honeycomb::Rack)
   require "rack/test"
-  require "rack/lobster"
+  begin
+    require "rack/lobster"
+  rescue LoadError
+    require "rack/session"
+    require "rackup"
```
emilyashley, line 9: on second thought, I'm not sure if this line is needed. @robbkidd ?

robbkidd: 🤔 I think you may be right. I will check!

emilyashley: It wasn't ✨ Just pushed a commit removing the duplicate "rackup" and also adding the rack-session dep to the appraisals file (not just directly in the .gemfile for rack_3)

JamieDanielson (commented):
Nice work! I'm not super familiar with Ruby, so I'll add my notes here from trying things out with this PR:

These changes worked for running `bundle exec appraisal rake test`. This makes sense - it worked in Circle! But when I tried `bundle install`, it was failing with this error:

`[!] There was an error parsing ``rack.gemfile``: No such file or directory @ rb_sysopen - /Users/jamiedanielson/code/beeline-ruby/gemfiles/rack.gemfile. Bundler cannot continue.`

I believe this was actually because of the rack example app, which should maybe be updated with this change? In `examples/rack/Gemfile` I added `gem "rackup"`; in `examples/rack/main.rb` I added `require "rackup"` to the top, and updated the handler to `handler = Rackup::Handler::WEBrick` (previously `Rack::`).

Now I can run `bundle install` successfully, and from `examples/rack/` I can run `bundle exec ruby main.rb`, which starts the server. Then I `curl localhost:8080` to get a`Hello from Honeycomb` reply, and see trace data in the console `{"meta.beeline_version":"2.11.0","meta.local_hostname":"jamiedanielson","meta.instrumentations_count":10,"meta.instrumentations":"[\"active_support\"...`

examples/rack/Gemfile
```diff
@@ -4,5 +4,6 @@ source "https://rubygems.org"
 
 git_source(:github) {|repo_name| "https://github.com/#{repo_name}" }
 
-gem "rack"
+gem "rack", ">= 3.0"
+gem "rackup"
```
robbkidd, line 8: This is what it took to fix the rack example now that it was pulling down Rack v3.

@JamieDanielson, your local environment still looking at `gemfiles/rack.gemfile` seems like the settings Appraisal manipulates for the different named-sets of dependencies are in effect for the old "rack" scenario which—yea—will make everything else in the repo very confused.

spec/honeycomb/integrations/rack_spec.rb
```diff
@@ -1,31 +1,41 @@
...
+  end
   require "rack/test"
-  require "rack/lobster"
   require "warden"
 
+  class AnApp
+    def call(_env)
+      [200, { "content-type" => "text/plain" }, ["Hello world!"]]
+    end
+  end
```
robbkidd, line 15: Huge fan of [replacing Rack::Lobster](https://github.com/honeycombio/beeline-ruby/pull/208/files#diff-9a936bf6526ca7d7934829e5fe3999ba63ad4e40a1160c586586b65d7d6d8401R15-R19)—as cute at it is—with a simple test app of our own to reduce dependencies.

spec/honeycomb/integrations/rack_spec.rb
```diff
@@ -1,31 +1,41 @@
...
+  # Rack::Session was moved to a separate gem in Rack v3.0
+  if Gem::Version.new(::Rack.release) >= Gem::Version.new("3.0")
+    require "rack/session"
+  end
   require "rack/test"
-  require "rack/lobster"
   require "warden"
 
+  class AnApp
+    def call(_env)
```
robbkidd, line 12: Starting the unused variable with an underscore will appease the Lint/UnusedMethodArgument Rubocop cop.

robbkidd:
> Longer secret required for Rack 3
— @emilyashley 

Ooops

spec/honeycomb/integrations/rack_spec.rb
```diff
@@ -1,31 +1,41 @@
...
+      Honeycomb::Rack::Middleware.new(base_test_app, client: client)
     end
     let(:auth) { Authenticate.new(honeycomb) }
     let(:warden) do
       Warden::Manager.new(auth) do |manager|
         manager.default_strategies :test
       end
     end
-    let(:session) { Rack::Session::Cookie.new(warden, secret: "honeycomb") }
+    let(:secret) { "honeycombsecretneedstobe64characterslongforrack3sonowitslongeryay" }
```
robbkidd, line 37: 🎉

robbkidd (approved):
It wasn't a rock! It was a rack lobster!

spec/honeycomb/integrations/rack_spec.rb
JamieDanielson: TIL

JamieDanielson (approved):
thats amazing

PR #218, opened 2023-01-03
fix: set min rubocop to avoid false positive lint error

robbkidd:

[v0.60.0](https://github.com/rubocop/rubocop/blob/master/CHANGELOG.md#0600-2018-10-26) includes a fix (6331) for false positives on Regexes assigned to a constant that will quiet the lint error that's appeared in nightly.

Less than v0.69.0 remains for Ruby 2.2 until a time that we decide to overhaul and update dev/test.


JamieDanielson (approved):
lint

PR #219, opened 2023-01-06
ci: use context for publishing ruby gem

JamieDanielson:

- use circle context for publishing ruby gem
- use `RUBYGEMS_API_KEY` instead of `GEM_HOST_API_KEY`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -291,6 +291,7 @@ workflows:
           filters: *tag_filters
           requires:
             - build_artifacts
+          context: Honeycomb Secrets for Public Repos
       - publish_github:
           <<: *publish
           context: Honeycomb Secrets for Public Repos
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -291,6 +291,7 @@ workflows:
           filters: *tag_filters
           requires:
             - build_artifacts
+          context: Honeycomb Secrets for Public Repos
```
robbkidd, line 294: With the added to this `publish_rubygems` YAML key that is aliased as `&publish`, the `context:` entry could be removed from the following `publish_github` section which includes all of `&publish`.

robbkidd (approved):
I have one suggestion, but I don't think it _needs_ to be done before merge.

.circleci/config.yml
JamieDanielson: ooh good eye! https://github.com/honeycombio/beeline-ruby/pull/219/commits/d44b676a5ed952ac1ee6e401952962798f3f7065

robbkidd (approved):
look at the context

PR #225, opened 2023-06-01
docs: update readme

vreynolds:
A few more words and visuals about beeline status


JamieDanielson (approved):
Failing build is the same as the nightly fail on older versions that we are investigating separately.

robbkidd (approved):

MikeGoldsmith (approved):
Ignoring build errors as they is updating README only.

PR #232, opened 2023-10-02
fix: clean invalid UTF8 from trace field values when generating propagation header

### LINKED ISSUE #231: Trace-level fields don't get "cleaned" for header serialization
Noticed an inconsistency between the data that gets sent by Libhoney on the transmission layer vs what gets serialized into the trace header.

You can add an invalid utf-8 string to a libhoney event, and it is saved directly in a hash table: https://github.com/honeycombio/libhoney-rb/blob/a2552e822d0bf27e3d599e6e50668e9983e05b5e/lib/libhoney/event.rb#L61-L88

When the event is sent by the default transmission, that data gets munged on its way to being JSON-encoded by `Libhoney::Cleaner`: https://github.com/honeycombio/libhoney-rb/blob/a2552e822d0bf27e3d599e6e50668e9983e05b5e/lib/libhoney/transmission.rb#L266 -> https://github.com/honeycombio/libhoney-rb/blob/a2552e822d0bf27e3d599e6e50668e9983e05b5e/lib/libhoney/cleaner.rb

However, when you call `Honeycomb::Span#to_trace_header`, the "raw" trace-level fields are taken from the `Honeycomb::Trace#fields` and passed directly into `JSON.generate`: https://github.com/honeycombio/beeline-ruby/blob/c7353316b47fb773fb79e8d805fd6aa9a20737a1/lib/honeycomb/propagation/honeycomb.rb#L62

Without the `Libhoney::Cleaner` acting on the data, that means the trace header serialization blows up with an error on improperly-encoded strings, whereas the span itself can be sent to Honeycomb just fine.

**Versions**

- Ruby: `ruby 3.2.2 (2023-03-30 revision e51014f9c0) [x86_64-darwin22]`
- Beeline: `honeycomb-beeline (2.11.0)`

**Steps to reproduce**

```ruby
require 'honeycomb-beeline'

Honeycomb.configure { |config| config.debug = true }

span = Honeycomb.start_span(name: 'span')
span.add_trace_field('invalid_utf8', "\xAE")

begin
  span.send
  puts "Sending works fine, due to Libhoney::Cleaner#clean_string"
rescue => e
  puts "Actually, Honeycomb::Span#send didn't work: #{e}"
end

begin
  header = span.to_trace_header
  puts "Actually, Honeycomb::Span#to_trace_header worked: #{header}"
rescue => e
  puts "Honeycomb::Span#to_trace_header blew up, because the trace-level fields don't get cleaned: #{e}"
end
```

```console
$ ruby /tmp/repro.rb
{"meta.beeline_version":"2.11.0","meta.local_hostname":"Alexs-MacBook-Pro.local","meta.instrumentations_count":10,"meta.instrumentations":"[\"active_support\", \"aws\", \"faraday\", \"rack\", \"rails\", \"railtie\", \"rake\", \"redis\", \"sequel\", \"sinatra\"]","service_name":null,"service.name":null,"name":"span","duration_ms":0.05299970507621765,"trace.trace_id":"957ed21fa9cd74f2808e11088f080cea","trace.span_id":"ec9ad6134bd1b0a2","meta.span_type":"root","invalid_utf8":"�"}
Sending works fine, due to Libhoney::Cleaner#clean_string
Honeycomb::Span#to_trace_header blew up, because the trace-level fields don't get cleaned: source sequence is illegal/malformed utf-8
```

**Additional context**

We were noticing this behavior in particular because we add the User-Agent header as a trace-level field. Spans normally get sent fine: the User-Agent field gets mangled with � characters (as can be seen in the repro). But on code paths where we make distributed calls and try to inject the trace header on outbound requests, we were triggering a hard error from `#to_trace_header` that interrupted the whole request cycle.

robbkidd:

- Resolves #231 

### Tests to Demonstrate the Problem

* Tests added to cover invalid UTF8 characters in trace field values for classic and E&S modes of Honeycomb propagation

### MarshalTraceContext

* Replace invalid UTF8 characters when serializing in Honeycomb-flavored propagators

Mixes in the data cleaning logic already present in Libhoney::Cleaner to prepare trace data for JSON serialization.

* Refactor the encoding logic for serialization in an extracted method to avoid having the two implementations diverge.

### UnmarshalTraceContext

* Decode trace context with the decoder that matches the encoder.

Testing revealed that decoding trace context would fail to produce a string parseable as JSON if some UTF8 characters were included in the data, even if those characters are valid. Because we encode with urlsafe, we need to decode with urlsafe.

### Linting

* Disable the Style/AccessModifierDeclarations Rubocop rule.

Sometimes we group methods under "private", sometimes we inline a symbol reference to a method with "module_method". We remain flexible. This rule does not serve us and we do not serve it.

### DIFF
--- a/lib/honeycomb/propagation/honeycomb.rb
+++ b/lib/honeycomb/propagation/honeycomb.rb
@@ -65,14 +65,11 @@ module Honeycomb
       module_function :clean_data, :clean_string
 
       def to_trace_header
-        fields = clean_data(trace.fields)
-        context = Base64.urlsafe_encode64(JSON.generate(fields)).strip
-        encoded_dataset = URI.encode_www_form_component(builder.dataset)
         data_to_propogate = [
-          "dataset=#{encoded_dataset}",
+          "dataset=#{encode_dataset(builder.dataset)}",
           "trace_id=#{trace.id}",
           "parent_id=#{id}",
-          "context=#{context}",
+          "context=#{encode_trace_fields(trace.fields)}",
         ]
         "1;#{data_to_propogate.join(',')}"
       end
@@ -84,18 +81,28 @@ module Honeycomb
       end
 
       def self.to_trace_header(propagation_context)
-        fields = clean_data(propagation_context.trace_fields)
-        context = Base64.urlsafe_encode64(JSON.generate(fields)).strip
-        dataset = propagation_context.dataset
-        encoded_dataset = URI.encode_www_form_component(dataset)
         data_to_propogate = [
-          "dataset=#{encoded_dataset}",
+          "dataset=#{encode_dataset(propagation_context.dataset)}",
           "trace_id=#{propagation_context.trace_id}",
           "parent_id=#{propagation_context.parent_id}",
-          "context=#{context}",
+          "context=#{encode_trace_fields(propagation_context.trace_fields)}",
         ]
         "1;#{data_to_propogate.join(',')}"
       end
+
+      def encode_trace_fields(fields)
+        Base64.urlsafe_encode64(
+          JSON.generate(
+            clean_data(fields)
+          )
+        ).strip
+      end
+      module_function :encode_trace_fields
+
+      def encode_dataset(dataset)
+        URI.encode_www_form_component(dataset)
+      end
+      module_function :encode_dataset
     end
   end
 end
--- a/lib/honeycomb/propagation/honeycomb.rb
+++ b/lib/honeycomb/propagation/honeycomb.rb
@@ -3,6 +3,7 @@
 require "base64"
 require "json"
 require "uri"
+require "libhoney/cleaner"
 
 module Honeycomb
   # Parsing and propagation for honeycomb trace headers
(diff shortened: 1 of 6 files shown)


spec/honeycomb/propagation/honeycomb_modern_spec.rb
```diff
@@ -75,6 +75,18 @@
       nil,
     ]
   end
+
+  it "handles previously-encoded invalid utf8" do
```
ajvondrak, line 79: Ah, good thinking! Hadn't considered this part of the "changing airplane wings mid-flight". ✈️

lib/honeycomb/propagation/honeycomb.rb
```diff
@@ -58,8 +59,14 @@ def parse_v1(payload)
 
     # Serialize trace headers
     module MarshalTraceContext
+      # for cleaning data in trace fields before serializing to prop header value
+      include Libhoney::Cleaner
+      # promote cleaner instance methods to module methods so that self.to_trace_header can use them
+      module_function :clean_data, :clean_string
```
ajvondrak, line 65: As I was monkey-patching our code locally, I was lamenting that `Libhoney::Cleaner` didn't already make these `module_function`s, so that I could just call `Libhoney::Cleaner.clean_data` as a singleton. 😛 But yeah, this looks about how I expected it to look. (Our monkey-patch goes through `add_trace_field` instead, cuz it was marginally easier to hack around, but I do think the cleaning properly belongs in the propagation code. 👍)

ajvondrak (commented):
Thanks, @robbkidd! Saves me the PR-slinging. 😄 And looks to have the shape I would expect (though admittedly I'm not scouring this PR for edge cases).

lib/honeycomb/propagation/honeycomb.rb
ajvondrak: Yeah, totally. I'd do the same. Fewer gems to bump, lol.

spec/honeycomb/propagation/honeycomb_modern_spec.rb
robbkidd: I'd love to take credit for being super smart and thinking of adding this case ahead of time, but really I was guided to add this because the fix for your original Marshal issue revealed [a bug in the Unmarshal implementation](https://github.com/honeycombio/beeline-ruby/pull/232/commits/3dcc5f60b8a28153ae0755af9321276c9073ecb2).

MikeGoldsmith (approved):

PR #238, opened 2024-03-06
feat: support classic ingest keys

robbkidd:

- Supporting ingest keys in classic environments

- [x] Consolidated the beeline's classic logic to the Configuration object.
- [x] Spec'd the classic logic directly for the configuration.classic? method and updated other specs to use a stub for classic/not-classic
- [x] Spec'd the new ingest key behavior (which fails)
- [x] Update minimum version of libhoney dependency to [the release](https://github.com/honeycombio/libhoney-rb/releases/tag/v2.3.0) that contains the new [public `Libhoney.classic_api_key?` method](https://github.com/honeycombio/libhoney-rb/pull/138) and use that method for key checking

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206669928045914

### DIFF
--- a/lib/honeycomb/client.rb
+++ b/lib/honeycomb/client.rb
@@ -33,7 +33,7 @@ module Honeycomb
       @libhoney.add_field "service.name", configuration.service_name
       @context = Context.new
 
-      @context.classic = classic_write_key?(configuration.write_key)
+      @context.classic = configuration.classic?
 
       @additional_trace_options = {
         presend_hook: configuration.presend_hook,
@@ -127,10 +127,6 @@ module Honeycomb
       )
       span.add_field("error_backtrace_limit", error_backtrace_limit)
       span.add_field("error_backtrace_total_length", exception.backtrace.length)
-    end
-
-    def classic_write_key?(write_key)
-      write_key.nil? || write_key.length == 32
     end
   end
 end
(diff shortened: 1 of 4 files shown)


lib/honeycomb/client.rb
```diff
@@ -128,9 +128,5 @@ def add_exception_data(span, exception)
       span.add_field("error_backtrace_limit", error_backtrace_limit)
       span.add_field("error_backtrace_total_length", exception.backtrace.length)
     end
-
-    def classic_write_key?(write_key)
```
robbkidd, line 132: This was a private method and its only use was to set the `classic` instance variable on `@context`. That is set by asking the `configuration` object directly.

spec/honeycomb/configuration_spec.rb
```diff
@@ -43,8 +43,41 @@
     expect(configuration.write_key).to eq write_key
   end
 
+  describe "#classic?" do
```
robbkidd, line 46: Spec the classicness of the various key formats once here.

spec/honeycomb/configuration_spec.rb
```diff
@@ -61,10 +94,12 @@
 
     context "with a classic write key" do
       before do
-        configuration.write_key = "c1a551c000d68f9ed1e96432ac1a3380"
+        allow(configuration).to receive(:classic?).and_return(true)
```
robbkidd, line 97: Stub the classic check within "classic" test scopes instead of repeating the matrix of key types all over again.

jharley (approved):
💎

MikeGoldsmith (approved):

PR #240, opened 2024-03-07
rel: 3.1.0

jharley:


CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # beeline-ruby changelog
 
+## 3.1.0 2024-03-07
+
+### Enhancements
+
+- feat: support classic ingest keys (#238) | Robb Kidd
```
jharley, line 7: Unclear if we want to put a note in about the min. libhoney version bump?

robbkidd: Probably a good idea. I don't _expect_ that to hurt anyone. 🤞🏻 😅 🤞🏻  

Maybe as a note for the release above the **Enhancements** section?

jharley: Done 👍🏻

CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
 # beeline-ruby changelog
 
+## 3.1.0 2024-03-07
+
+⚠️  Minimum libhoney version is now 2.3.0 ⚠️
+
```
robbkidd, line 6:
```suggestion
⚠️  Minimum libhoney version is now 2.3.0 ⚠️

Previously the minimum libhoney was 1.14.2. Backwards-incompatible changes in libhoney 2.0.0:

* Removal of the deprecated Array input for http proxy configuration.
  * If you are using the Beeline in services that must send web traffic through an out-going proxy, see [libhoney 1.x proxy_config deprecation warning](https://github.com/honeycombio/libhoney-rb/blob/v1.21.0/lib/libhoney/client.rb#L273-L276) for recommended configuration.
* Drop support for Ruby versions 2.2 and 2.3.
  * Beeline Ruby already requires a minimum Ruby 2.5.

 Beeline Ruby already has a minimum Ruby version of 2.5.
 
```

robbkidd (approved):
👍🏻 

I'm an over-informer with major version warnings like this, so I dropped a suggested addition to this release note.

PR #245, opened 2025-01-23
maint: Add link to sunset issue in README

MikeGoldsmith:

Add link to sunset issue in README.


robbkidd:
We'll also need to update the value in OSSMETADATA and the color of the oss-lifecycle badge at the top of the README.

